Issue #25, opened 2016-05-18
umi_tools extract run error


royfrancis:
I have paired end data (2 fastq files, 77 mil reads) and I have 8 nt UMI bases on the 5' end of the first fq file.

I have a shell script file with this content:
 
```Bash
newfilename="${1/_*/}"

cat "$1" | umi_tools extract \
--bc-pattern="NNNNNNNN" \
--read2-in="$2" \
--read2-out="$newfilename-um_2.fq" \
-L "umitools-extract.log" > "$newfilename-um_1.fq"
```
where  `$1` is the first fq file (fwd) and `$2` is the second fq file (rev). And run as `script.sh file_1.fq file_2.fq`. I get the below error after the script runs for a couple of minutes.
```
sbatch-umitools-extract.sh: line 42: 32582 Broken pipe             cat "$1"
     32583 Killed                  | umi_tools extract --bc-pattern="NNNNNNNN" --read2-in="$2" --read2-out="$newfilename-um_2.fq" -L "umitools-extract.log" > "$newfilename-um_1.fq"
```

Strangely, it works fine on a small test dataset (paired-end fastq) of 40 lines each. The output fq files show the UMIs have been correctly removed and added to the read name. But it doesn't work for my big dataset. I wonder if it is just a script syntax error or something else.

IanSudbery:
Hi There,

I think this is almost certainy due to running out of memory - the "Killed" suggests that the job was killed by the OOM killer. The broken pipe comes from the fact that once the umi_tools process died, the cat had nowhere to send data. 

I'm going to have a look at why it might be using so much memory.

royfrancis:
I had tried increasing the memory from 16GB to 48GB and I still get the same error.

IanSudbery:
Hi Roy,

Can you try the code in [extract_memory](/CGATOxford/UMI-tools/tree/extract_memory) branch. It should reduce the memory usage from 100s of GB to 100s of MB for paired end files. 

Let me know if this fixes your problem.

royfrancis:
Hi Ian,
This time, it worked, the run completed successfully without errors in 30 mins (74 mil reads paired-end). Available resources was 6 cores and 48gb RAM. This was reported from the job controller during the run:
`slurmstepd: Exceeded step memory limit at some point.`
which usually means that the job had exceeded the pre-allocated 48gb.

But on inspecting the job report, it seems like only a single core was used and RAM usage was low. Perhaps this may not be reliable. 

[...]

PR #26, opened 2016-05-20
Use iterator to reduce memory usage in paired end processessing

IanSudbery:
Fixes issue #25

### DIFF
--- a/umi_tools/extract.py
+++ b/umi_tools/extract.py
@@ -82,9 +82,8 @@ Command line options
 --------------------
 
 '''
-
 import sys
-
+from itertools import izip
 import umi_tools.Utilities as U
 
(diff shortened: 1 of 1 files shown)


IanSudbery:
Just checking this is good with you @TomSmithCGAT

TomSmithCGAT:
Great. Thanks!

TomSmithCGAT:
@IanSudbery I've added in a try, except to make izip python 2 & 3 compatible feda398 . Otherwise we get

``ImportError: cannot import name 'izip'``

IanSudbery:
Ah yes! zip is an iterator by default in py3. Perhaps we should have py3 build tests as well as py2?

Issue #29, opened 2016-05-23
UMI removal


casimonet:
Hi,
Sorry to open another issue. The fact that the UMI is removed from the sequence is a problem for downstream analyses. For example for building a bam file where the reads are filtered for UMI in which bases have a low quality. Or simply to keep a column with the UMI in the bam file, which can be more practical than having the UMI in the reads name for certain downstream analyses.

Is there no way to avoid the read to be truncated?

Thank you,
Camille

TomSmithCGAT:
Hi Camile,

No need to apologise. Comments are always very welcome.

You raise two issues:

1. Filtering on quality
2. Retaining the UMI elsewhere in the BAM file

1. Filtering on quality:
It's not realistically possible to cover all possible filtering one might wish to apply to the reads. Previously, where I've needed to do this, I've written bespoke code to perform the filtering at the UMI extraction stage as this step is normally relatively simple. I can see that filtering on quality would be a useful feature in general though and easy to implement so I'm happy to add it to the extract tool.

2. For the alignment step it's necessary to retain the UMI in the read ID as this ensures all aligners will retain the UMI sequence in the resultant BAM. I guess it would be possible to have an option within dedup to move the UMI to a BAM field during the de-deduplication. Can you give me a motivating example as to why this would be useful?

Regards,

Tom

Tom Smith, PhD
CGAT Training Fellow | http://www.cgat.org<https://owa.nexus.ox.ac.uk/owa/redir.aspx?C=tg_bDY9IpU-WsZT9DRtecLrzBQBo5dAIR2VtNMFUhhmiSbB621cWM1yWzPrEKkt2W6Kx2kZl0Lg.&URL=http%3a%2f%2fwww.cgat.org>
MRC Functional Genomics Unit | University of Oxford
+44 1865 285854 | Thomas.Smith2@dpag.ox.ac.uk
https://cgatoxford.wordpress.com/author/tss505/
________________________________

IanSudbery:
Retaining the UMI on the read is not possible: if you were to leave the UMI on the read, your reads would not map. 

I think adding quality filtering to extract is probably a good idea in the medium to long term. However, until that point i'd recommend filtering on the UMI quality before trimming the UMI.

[...]

PR #33, opened 2016-05-25
Ts add quality filtering

TomSmithCGAT:
Implements quality score filtering for UMI bases.
See #29 

@IanSudbery Would you mind checking this.


IanSudbery:
Looks fine, but we should make sure we test it on single and paired-end files, with UMIs on the 3' and 5' end.

TomSmithCGAT:
Agreed. I'll do this tomorrow.

TomSmithCGAT:
@IanSudbery Checked on single/paired, 3'/5' and all seems OK. Updated now so that only the UMI bases from the pattern are checked for quality. Previously all positions in the bc-pattern were checked.

PR #36, opened 2016-06-02
Ts speed up stats

TomSmithCGAT:
@IanSudbery After discussing with @MikeDMorgan I've implemented a two-stage random sampling since the original array of probabilities contains the same probability multiple times (e.g 1/n, 2/n, 3/n, ..., where n is the total umi counts) This speeds up the random sampling ~5-fold for the BAM from Roy Francis (see below) so the total run time decreased from ~28 minutes to ~10 minutes. This may be a particular case where the gains are larger since the distribution of UMIs is a long way off uniform. I'd expect the improvement to be marginal where the distribution of umis is more "random".

I've checked the stats output is consistent. 

Can you think of any scenario where this method of sampling would be significantly less efficient?

previously...
ncalls  tottime  percall  cumtime  percall filename:lineno(function)
  4400216 1373.266    0.000 1426.919    0.000 {method 'choice' of 'mtrand.RandomState' objects}
  4400216   18.888    0.000 1445.807    0.000 dedup.py:776(getUmis)

two-stage sampling...
  8995437  219.319    0.000  268.225    0.000 {method 'choice' of 'mtrand.RandomState' objects}
  4400216   24.821    0.000  300.054    0.000 dedup.py:789(getUmis)

### DIFF
--- a/umi_tools/dedup.py
+++ b/umi_tools/dedup.py
@@ -176,12 +176,17 @@ import pysam
 import pandas as pd
 import numpy as np
 
-import umi_tools.Utilities as U
+try:
+    import umi_tools.Utilities as U
+except:
+    import Utilities as U
 
 import pyximport
 pyximport.install(build_in_temp=False)
-from umi_tools._dedup_umi import edit_distance
-
+try:
+    from umi_tools._dedup_umi import edit_distance
+except:
+    from _dedup_umi import edit_distance
 
 def breadth_first_search(node, adj_list):
     searched = set()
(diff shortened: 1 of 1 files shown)


IanSudbery:
Hi Tom,

I had always thought it was the reading in of the UMIs that was slow, rather than the actual random sampling! The only way I can see this being less efficient is if there is only one UMI in each frequency bin, and thus the actual random choice become equivalent, but now you've the added burden of creating the frequency2umi dict. But I think even here, it shouldn't be too much of a hit because its a one off cost.

IanSudbery:
memory usage looks good. I recommend merging.

TomSmithCGAT:
Thanks @IanSudbery

PR #41, opened 2016-08-08
Compatibility with python3

peterch405:
Didn't test with python2


IanSudbery:
Tested in python 2 - runs fine.

PR #53, opened 2016-10-31
Install

TomSmithCGAT:


IanSudbery:
I can't say I understand what is going on with the travis stuff. In working out the installation instruction, I made sure that it *wasn't* neccessary to install cython before installing umi_tools. The only reason it would be neccesary would be if you were building either pandas or pysam from source, which shouldn't be happening here - they should be being installed from wheels.

TomSmithCGAT:
I think there were a variety of problems depending on the commit we're talking about. I think the cython issue was that whenever travis finds a _requirements.txt_ it runs ``pip install -r requirements.txt``, hence the CGAT requirements being contained in _requires.txt_. I think this was installing pandas before cython was installed which causes the cython-related fail. There were multiple changes though so I may have misremembered. 

Regardless, the travis.yml now uses the commands correctly to explicity set up the correct environment. We're using _before_install:_ to install cython with pip, then the _install:_ to install umi_tools. The tests are simply that the _--help_ commands can be run. We still need to implement a proper test script.

PR #56, opened 2016-11-09
Tests

TomSmithCGAT:


TomSmithCGAT:
@IanSudbery We now have automated testing. See  tests/tests.yaml for a list of tests and suggestions for further tests that are required

IanSudbery:
Great!

TomSmithCGAT:
Note that one of the tests sometimes need re-running in travis due to a failure at the `python setup.py install` step. I've not yet worked out why the test sometimes fails here. The two tests are run in the exact same environment so this is a bit of a puzzle?

See https://travis-ci.org/CGATOxford/UMI-tools/builds/174537559 for an example.

Issue #54, opened 2016-10-31
Output read groups rather than deduplicated BAM


TomSmithCGAT:
@IanSudbery Just had a chat with Nils Koelling here at the WIMM. He's given some very helpful feedback on documentation which I'll feed into the master branch shortly. 

He's using UMIs in a variant calling framework and therefore wants to group reads by their UMI and then perform a downstream step to identify the correct base call for each position in the set of reads to correctly resolve sequencing errors. We spoke about a couple of ways to achieve this:

1. Add a tag to the BAM with a unique group identifier (1 -> infinity)
2. Output a csv mapping read identifier to group

Theoretically, these could be achieved within the dedup command but this is getting away from a pure deduplication problem as such. I think the best approach might be to provide a new umi_tools command _"group"_ which uses the same network methods to identify the set of reads with the same UMI or related UMIs and then outputs either a tagged BAM or csv. We'd need to abstract some of the dedup.py code to a module file and reuse it in group.py.

This might also be a good way to address this issue regarding counting reads supporting each UMI: https://github.com/CGATOxford/UMI-tools/issues/45

What do you think?

IanSudbery:
I've been wondering recently why no one does this! I think we probably might want to do this in a new command - otherwise we are going to have to store all of the reads in memory - at the moment we only store the best of each position*splice*umi combo. At high read depths this is a substantial memory saving. 

However, the new problem is kind of a deduping problem though? Presumably you only want to group together reads with the same UMI and same position.

I'd be interested to see how this progresses - I've got some project where using duplicates to resolve sequencing errors could be very useful.

TomSmithCGAT:
Yeah, in this case this is ultimately a de-duplication problem but the intermediate step could be quite bespoke, hence the solution to just return the read groups and allow the user to decide how to use the information. 

I'll make a group branch and start work shortly.

koelling:
Great, thank you very much Tom! Happy to help out with testing etc!

TomSmithCGAT:
@koelling Would you mind testing out the new group command (on the group branch). Latest commit here 9579cad5f4e5f0f6c437605d2f70bbb2fe27eb5a. You can run it like so:

``umi_tools group -I example.bam  --group-out=test.tsv``

Currently, it only outputs a tsv mapping read_identiers to read groups according to their position and UMI. Once I've figured out how to add a tag to the reads, the following command should also generate a bam with the read group included as the 'UG' tag.

[...]

PR #57, opened 2016-11-10
Group

TomSmithCGAT:
@IanSudbery 

The group branch includes a new command "group" which will attempt to identify the group of UMIs which come from the same True UMI + errors and output either a tagged BAM and/or a flatfile detailing the UMI groups.

To achieve this, much of the code originally in umi_tools/dedup.py was refactored into two module files, network.py and umi_methods.py (this could do with a better name). In additional, the functionality of the twopasswriter method was expanded to allow addition of tags to reads and the functionality of the ClusterAndReducer functor (renamed ReadClusterer) was expanded to allow the use of the functor for grouping as well as deduplicating.

I'll leave the merging until you've had a chance to look over the code. There is no urgency to merge back into master immediately.


TomSmithCGAT:
@IanSudbery Closed until #54 resolved

TomSmithCGAT:
@IanSudbery For the time-being I've removed the option to use adjacency with the group command until I've resolved the issue with some read being reported multiple times (see #54).

PR #62, opened 2016-12-02
Tom smith cgat patch setuptools

TomSmithCGAT:
@IanSudbery I've hardcoded a minimum requirement for setuptools in the use_setuptools() function. previously we didn't specify a minimum version so it defaults to the latest version. 

I think this is the most likely explanation for the odd installation errors where it appears to reinstall setuptools even after installing all the requirements. This was stopping the conda build working for me. However, this also suggests the easy install solution might not work when required?

### DIFF
--- a/ez_setup.py
+++ b/ez_setup.py
@@ -157,7 +157,6 @@ def use_setuptools(
 
     try:
         import pkg_resources
-        print "`\n\n\n\n\n\n\nsetuptools>=" + version
         pkg_resources.require("setuptools>=" + version)
         # a suitable version is already installed
         return
--- a/ez_setup.py
+++ b/ez_setup.py
@@ -157,6 +157,7 @@ def use_setuptools(
 
     try:
         import pkg_resources
+        print "`\n\n\n\n\n\n\nsetuptools>=" + version
         pkg_resources.require("setuptools>=" + version)
         # a suitable version is already installed
         return
--- a/setup.py
+++ b/setup.py
@@ -2,8 +2,8 @@ import sys
 import os
 import glob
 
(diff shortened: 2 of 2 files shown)


IanSudbery:
Hi Tom,

Can you hold off on merging this until I've tried running it on a clean testing VM to check if the most low level installation instructions still work?

Ian

TomSmithCGAT:
@IanSudbery Sure. The only impact should be that it won't run the _do_download function in ez_setup to download the latest setup tools.

IanSudbery:
Hi Tom,

This branch does not succesfully install from source on a clean Ubuntu VM, but fails with error message:
numpy/core/src/npymath/ieee754.c.src:7:29: fatal error: npy_math_common.h: No such file or directory

I've check 0.2.6 again and that still works, so its not to do with updates in upstream dependencies. I'm going to check 0.3.1 so see if that works.

TomSmithCGAT:
Is this relevant?
http://abriefsummaryofeverything.blogspot.co.uk/2016/08/ieee754csrc729-fatal-error.html

IanSudbery:
Well that is the error, but I am not using virtualenv.

TomSmithCGAT:
@IanSudbery Any progress on this? I want to merge in these changes before I update the version available on conda

IanSudbery:
@TomSmithCGAT Some tinkering has revealed that we require setuptools>= 10.0 in order for setup to run.

TomSmithCGAT:
@IanSudbery Great. Thanks. Are you happy to merge this into master?

IanSudbery:
Yes. Go ahead

PR #65, opened 2017-01-18
Py3.5 testing

TomSmithCGAT:
Resolves errors when using python3.

- Fastq iterator returns bytes not string with Py3. Implemented conversion to str. Is there a better way to deal with this within the call to .readline()?
- Can't iterate over dictionary keys with Py3 so converted dict keys to list


TomSmithCGAT:
Note. The tests are failing for dedup and group due to the output order. See #66. If we can't get this sorted soon, I propose merging the changes regardless as the current version will not work with py3 at all and the inconsistency in output order between py2 and py3 is a minor issue.

TomSmithCGAT:
See #65 For discussion

Issue #73, opened 2017-02-01
Read UMI from tag


Poshi:
In this tool, you are assuming that the UMI will be found in the ID, but there can be cases where the UMI is in other places. In our case, the UMI is located in a tag. Specifically in the RX tag (but others use the BC/BX).
Also, when assigning the UMI, the tags are competing (you use FU, Picard uses MI).

* It would be nice to be able to specify the tags to be used for all of these operations.
* And it would be nice to be able to read the UMI from the given tag, not only from the ReadID.

IanSudbery:
Hi,

Reading using UMIs specified in different locations is on our to do list for the next release. 

Ian

PR #76, opened 2017-02-01
Extract UMI from tag & Specify output tag

TomSmithCGAT:
- Adds options to extract UMI from tag rather than read ID (#73)
- Standardises UMI tag with Picard when using group command and outputting BAM (#73)

@IanSudbery Could you take a look over before I merge this into master (assuming tests pass...)


umi_tools/umi_methods.py
```diff
@@ -30,13 +30,28 @@
...
+            raise ValueError(
+                "Could not extract UMI from the read ID, please"
+                "check UMI is encoded in the read name")
+
+    elif method == 'tag':
+        try:
+            return read.get_tag(tag).encode('utf-8')
+        except IndexError:
+            raise ValueError(
+                "Could not extract UMI from the read ID, please"
```
TomSmithCGAT: Error message should say "Could not extract UMI from the read tag, please [...]"

umi_tools/umi_methods.py
```diff
@@ -370,7 +386,7 @@ def fill(self):
             if read.is_read2:
                 continue
 
-            self.umis[get_umi(read)] += 1
+            self.umis[get_umi(read, umi_sep, umi_tag, get_umi_method)] += 1
```
TomSmithCGAT: Previously we were using the get_umi function without passing the user defined umi separator! This is an bug I introduced when I added the --umi-separator option.

umi_tools/group.py
```diff
@@ -403,7 +440,7 @@ def main(argv=None):
                         else:
                             # Add the 'UG' tag to the read
                             read.tags += [('UG', unique_id)]
-                            read.tags += [('FU', top_umi)]
+                            read.tags += [('BX', top_umi)]
```
TomSmithCGAT: This is inline with Picard Markduplicates with molecular barcodes which uses RX for raw UMIs and BX for "corrected UMIs" - It's not clear how they 'correct' UMIs but the release notes for 2.7.0 says 'UMI Aware Mark Duplicates with rudimentary error correction' 
https://github.com/broadinstitute/picard/releases

umi_tools/umi_methods.py
```diff
@@ -30,13 +30,28 @@
...
-    except IndexError:
-        raise ValueError("Could not extract UMI from read, please"
-                         "check UMI is encoded in the read name"
-                         "following the final '_' or run with --ignore-umi")
+def get_umi(read, sep="_", tag='RX', method='read_id'):
+    ''' extract the umi from either:
+    - read id using specified separator
+    - specified tag'''
+
+    if method == 'read_id':
```
IanSudbery: I think that running an `if` every time a read is passed is going to slow things down. Have you done a profile? When we went through our fit of optimising things last week, by the end of it "get_umi" was the most time consuming function on some files (although that could be the `.encode` call, which seems for some reason to be very slow). 

An alternative is to have two separate functions get_umi_id and get_umi_tag. In dedup.py we can have something like:

```python
if options.get_umi_method=="read_id":
    umi_getter = partial(methods.get_umi_read_id, separator=options.umi_separator)
elif options.get_umi_method=="tag":
   umi_getter = partial(methods.get_umi_tag, tag=options.umi_tag)
else:
   raise ValueError("Unknown umi extraction method")
```

And then pass umi_getter to the required functions.

umi_tools/group.py
Poshi: Hummm... I'm not really sure how works the BARCODE_TAG/READ_ONE_BARCODE_TAG/READ_TWO_BARCODE_TAG in Picard, but I think that the assigned UMI in PICARD is marked with MI:
```
ASSIGNED_UMI_TAG (String)	Tag name to use for assigned UMI Default value: MI. This option can be set to 'null' to clear the default value.
```
https://broadinstitute.github.io/picard/command-line-overview.html#UmiAwareMarkDuplicatesWithMateCigar

TomSmithCGAT: No I'm not sure about how Picard works with UMIs. I was taking this from this forum 
http://gatkforums.broadinstitute.org/gatk/discussion/7497/picard-markduplicates-with-umi
I'll run Picard and check the output...

umi_tools/umi_methods.py
TomSmithCGAT: @IanSudbery I did profile and it seemed to make no difference whatsoever. I should probably check that again though.

I was originally going to set the get_umi method in the umi_methods.get_bundles function but as it's also used to generate stats in dedup. I presume we can set umi_getter in dedup as above and pass it to the umi_methods.get_bundles function?

umi_tools/group.py
Poshi: Yes... well, the UmiAwareMarkDuplicatesWithMateCigar command is still in alpha stage. Until the barcodes get somehow standardized... anything can happen.
BTW, allowing the user to set a specific tag and giving something as default tag should solve any future change in the stantards.

TomSmithCGAT: Yeah this was the conclusion I came to as well.

TomSmithCGAT:
@IanSudbery I've implemented the option to read UMIs from the read tag as you suggested. Would you mind reviewing this before we merge

Issue #44, opened 2016-08-24
Dedup by per-gene 


TomSmithCGAT:
Currently, dedup can work on a per-position or per-contig basis.The per-contig method is implemented for single cell RNA-Seq methods such as e.g CEL-Seq where the position of the read on a transcript may be different for reads generated from the same initial molecule. However, this requires that the reads be aligned to a single contig, i.e a representative gene model. Ideally we would like to align to all transcripts and give dedup a file mapping contigs (transcripts) to meta-contigs (genes) and introduce a "per-meta-contig" method. 

This is straightforward to implement within UMI-Tools dedup but requires the bam to be sorted by a meta-contig feature which I don't think is possible with standard command line tools. We could of course make a new command umi_tools sort to perform the sort using pysam?

@IanSudbery - thoughts?
Biostars question on custom sam sort tool [ here](https://www.biostars.org/p/208599/)

IanSudbery:
Is it neccesery to sort the BAM before hand? Could we not just, for each gene, run a fetch with a list of contigs associated with each meta-contig? Something like:

```python
def metafetcher(bamfile, contig2metacontig):

    def _inner(metacontig):
          for contig in contig2metacontig[metacontig]:
              for read in bamfile.fetch(contig): 
                   read.contig = metacontig
                   yield read

    return _inner


if options.per_gene:
      fetcher = metafetcher(in_bam, gene2transcript)
else:
      fetcher = pysam.fetch

```

It would of course require an indexed bam as the input rather than being able to read off the stdin.

TomSmithCGAT:
Yup, that's much simpler!

inakicm:
Hi,

Thanks for UMI_tools, and sorry if this is a silly question:

I can not use the per-position method because we used a protocol where the position of the read on the genome may be different for reads generated from the same initial molecule. 
I made the mapping using HTseq-count, so after the mapping I get a TAG (XF) in the SAM file which gives me the gene name for each read. So instead of working with the per-position method, is it possible to simply give a TAG (which corresponds to the gene name) to the deduplication procedure? In this way, the UMI deduplication works directly on the gene and not on the position.

Thanks,
Iñaki

IanSudbery:
Hi Inaki,

[...]

PR #88, opened 2017-02-28
adds per-gene deduplication option

TomSmithCGAT:
See #44 for motivation

This allows per-gene de-duplication by addition of a 'metafetcher' along the lines suggested by @IanSudbery. Each metacontig is considered in turn and all reads aligned to the contigs are tagged with the metacontig name which becomes the alignment position in reads_dict.

I've done some naive testing by manually modifying a samfile but the code is not currently covered by any testing.

### DIFF
--- a/umi_tools/dedup.py
+++ b/umi_tools/dedup.py
@@ -205,6 +205,25 @@ very long (>14bp)
 --chrom (string)
       Only consider a single chromosome. This is useful for debugging purposes
 
+--per-contig (string)
+      Deduplicate per contig. All reads with the same contig will be
+      considered to have the same alignment position. This is useful
+      if your library prep generates PCR duplicates with non identical
+      alignment positions such as CEL-Seq. In this case, you would
+      align to a reference transcriptome with one transcript per gene
+
+--per-gene (string)
+      Deduplicate per gene. As above except with this option you can
+      align to a reference transcriptome with more than one transcript
+      per gene. You need to also provide --gene-transcript-map option
+
+--gene_transcript_map (string)
+      File mapping genes to transripts (tab separated), e.g:
+
+      gene1   transcript1
+      gene1   transcript2
+      gene2   transcript3
+
 -i, --in-sam/-o, --out-sam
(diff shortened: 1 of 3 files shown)


umi_tools/dedup.py
```diff
@@ -484,18 +517,26 @@ def main(argv=None):
...
+        if options.per_gene:
+            metacontig2contig = umi_methods.getMetaContig2contig(
+                options.gene_transcript_map)
+            inreads = umi_methods.metafetcher(infile, metacontig2contig)
+        else:
+            inreads = infile.fetch()
 
     for bundle, read_events in umi_methods.get_bundles(
-            infile,
-            read_events,
```
TomSmithCGAT: read_events was being passed to get_bundles for no apparent reason

umi_tools/dedup.py
```diff
@@ -484,18 +517,26 @@ def main(argv=None):
...
+            inreads,
             ignore_umi=options.ignore_umi,
             subset=options.subset,
             quality_threshold=options.mapping_quality,
-            paired=options.paired, chrom=options.chrom,
```
TomSmithCGAT: chrom option is now handled prior to get_bundles

umi_tools/umi_methods.py
```diff
@@ -182,11 +182,38 @@ def get_read_position(read, soft_clip_threshold):
...
+    return metacontig2contig
+
+
+def metafetcher(bamfile, metacontig2contig):
+    ''' return reads in order of metacontigs'''
+    for metacontig in metacontig2contig:
+        for contig in metacontig2contig[metacontig]:
+            for read in bamfile.fetch(contig):
+                read.tags += [('MC', metacontig)]
+                yield read
```
TomSmithCGAT: This seems safer than updating the contig as the metacontig will be an unexpected value according to the sam header so could get problems with the reference id and reference nams etc.

TomSmithCGAT:
@IanSudbery This branch was merged into the {TS}-RetainUnmappedReads branch to avoid conflicts later so the changes are probably best reviewed there prior to merging into master.

PR #89, opened 2017-03-01
updates gettag method to support 10X UMI tag encoding

TomSmithCGAT:
10X long ranger pipelines add a tag with the UMI (BX) but this is also appended with the 'GEM' group, e.g
AGAGAGAGTA-1

Quoting from: https://support.10xgenomics.com/genome-exome/software/pipelines/latest/output/bam

_'This number denotes what we call a GEM group, and is used to virtualize barcodes in order to achieve a higher effective barcode diversity when combining samples generated from separate GEM chip channel runs. Normally, this number will be "1" across all barcodes when analyzing a sample generated from a single GEM chip channel. It can either be left in place and treated as part of a unique barcode identifier, or explicitly parsed out to leave only the barcode sequence itself.'_

### DIFF
--- a/umi_tools/umi_methods.py
+++ b/umi_tools/umi_methods.py
@@ -45,7 +45,9 @@ def get_umi_tag(read, tag='RX'):
     ''' extract the umi from the specified tag '''
 
     try:
-        return read.get_tag(tag).encode('utf-8')
+        # 10X pipelines append a 'GEM' tag to the UMI, e.g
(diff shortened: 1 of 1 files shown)


TomSmithCGAT:
@IanSudbery Are you happy for me to merge this?

IanSudbery:
I left this comment on the other PR, but have you profiled this to see what the time hit is for files that do not have GEMs?

PR #90, opened 2017-03-01
{ts} retain unmapped

TomSmithCGAT:
I've merged in the changes from {TS}-DedupPerGene to avoid conflicts later

**Per gene:**
This allows per-gene de-duplication by addition of a 'metafetcher' along the lines suggested by @IanSudbery. Each metacontig is considered in turn and all reads aligned to the contigs are tagged with the metacontig name which becomes the alignment position in reads_dict.

I've done some naive testing by manually modifying a samfile but the code is not currently covered by any testing.

**Output unmapped**
Unmapped reads are now yielded immediately and written out without the tags for corrected UMI and read group identifier. For paired end, if either read is unmapped, both and written out without tags.

I've also slightly modified the yielding of read bundles (umi_methods.py, line 323). Previously, when the contig changed, some reads from the previous contig were retained in the read_dict. This meant some reads were retained in the reads_dict until the very end when read_dict is cleared (line 410).


TomSmithCGAT:
Hi @IanSudbery Would you mind giving this a look over before I merge?

umi_tools/umi_methods.py
```diff
@@ -45,7 +45,9 @@ def get_umi_tag(read, tag='RX'):
     ''' extract the umi from the specified tag '''
 
     try:
-        return read.get_tag(tag).encode('utf-8')
+        # 10X pipelines append a 'GEM' tag to the UMI, e.g
+        # AGAGSGATAGATA-1
+        return read.get_tag(tag).split("-")[0].encode('utf-8')
```
IanSudbery: How much time does doing this to every read, irrespective of whether it contains a GEM or not add?

umi_tools/umi_methods.py
```diff
@@ -351,9 +401,15 @@ def get_bundles(insam, read_events,
                     reads_dict[pos][key][umi]["read"] = read
 
     # yield remaining bundles
+    for x in reads_dict:
+        U.info(reads_dict[x])
+        for y in reads_dict[x]:
+            U.info(x)
+            U.info(y)
```
IanSudbery: Why this output? Seems like if there were quite a lot of reads on the last contig, and you are in per-contig or per-gene, this is going to output a whole lot of unneccesary crap

TomSmithCGAT: Sorry, that's some debugging log statements that should have been removed

umi_tools/umi_methods.py
TomSmithCGAT: I'll check

TomSmithCGAT: No difference. Relevant sections from profiling

Master:
```
67254379 function calls (67225655 primitive calls) in 31.555 seconds
[...]
2367021    1.767    0.000    4.693    0.000 ../umi_tools/umi_methods.py:33(get_umi_read_id)
```

{TS}-RetainUnmapped:
```
67254475 function calls (67225746 primitive calls) in 29.743 seconds
[...]
2367021    1.761    0.000    4.489    0.000 ../umi_tools/umi_methods.py:33(get_umi_read_id)

```

IanSudbery:
Cool, go ahead and merge

umi_tools/umi_methods.py
TomSmithCGAT: Profiled master vs {TS}-RetainUnmapped using the test input file for extraction from bam 
e.g:
`python -m cProfile -o profile_master.cprof ../umi_tools/dedup.py --stdin=chr19_tag.bam -L test_master.log --out-sam --random-seed=123456789 --method=directional --umi-tag=RX --extract-umi-method=tag > test.bam`

5 times for each branch. Total times for 55197 calls to umi_methods.get_umi_tag:
Master: 0.027-0.029s
{TS}-RetainUnmapped: 0.043-0.051s

For reference, the total run time was 2.1-2.3s and the 55197 calls to pysam.get_tag took 0.83-0.100. 

umi_get_tag is only ever going to be called a maximum of 2*Number of reads. For input with lots of reads but simple networks, this will add <2% to the run time. For input with complex networks, the impact on run time will be minimal.

PR #98, opened 2017-03-21
added sort to group directional so that correct representative UMI is…

IanSudbery:
… selected. Tested by comparing outputs minus the corrected UMI (BX tag in sam, 6th column in tsv). Thus the UG is the same even if the BX is different. Have replaced tests


TomSmithCGAT:
Would you like me to update the py3 tests?

IanSudbery:
I completely forgot that python3 had separete tests. If you don't mind that would be great - I don't currently have a py3 environment to hand.

TomSmithCGAT:
No worries. I'll do this tomorrow afternoon and then merge.

PR #101, opened 2017-03-27
Do not cache read for mate output if mate is unmapped.

IanSudbery:
pysam.AlignedSegment.next_reference_name throws ValueError if the mate is unmapped.

If `group` is run with `--output-unmapped` then if read1 is mapped, but read2 is not, TwoPassWriter throws an error because of this. 

This fix skips trying to store the mate details for delayed outputting if the mate is unmapped. In this case the mate is output independently anyway.


TomSmithCGAT:
Good catch!

Issue #103, opened 2017-04-25
Headers for Sam output


TomSmithCGAT:
The option ` --out-sam` enables sam ouput. Previously, we did not include `h` in the output mode which defaulted to no header for sam output. From pysam 0.11 onwards, the default behaviour is now to include the header which does seem more sensible (https://github.com/pysam-developers/pysama/releases). 

This is causing all our tests to fail. I can resolve this by switching the header option off again for sam output but I think we should really include the header in the sam output and rebuild the test files. The header is apparently optional, however, I can't see an option to open a headerless sam with samtools (?) so our sam output is currently unusable.

This update would not be backwards compatible but I doubt this option is being used as nobody has raised the issue that the sam output does not contain a header and is not valid input for samtools. What do you think @IanSudbery?

IanSudbery:
I agree

TomSmithCGAT:
Fixed in {TS}-rebuild_tests

PR #104, opened 2017-04-26
py3 transparent gzipped output

IanSudbery:
Python3 requires an encoding and text mode to be specified to be able to output strings to a zip file. This was breaking the ability output to a zip file from extract if the user was using python 3.X.

This should really have tests, but I don't have time to implement right now.

### DIFF
--- a/umi_tools/Utilities.py
+++ b/umi_tools/Utilities.py
@@ -499,7 +499,16 @@ def openFile(filename, mode="r", create_dir=False):
             os.makedirs(dirname)
 
     if ext.lower() in (".gz", ".z"):
-        return gzip.open(filename, mode)
+        if sys.version_info.major >= 3:
+            if mode == "r":
+                return gzip.open(filename, 'rt', encoding="ascii")
+            elif mode == "w":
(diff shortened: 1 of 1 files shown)


IanSudbery:
The tests pass on my machine. @TomSmithCGAT can you check them on yours?

TomSmithCGAT:
Sure. I'll get on this tomorrow

TomSmithCGAT:
These fails are due to the pysam update and sam headers (#103). If you update to the latest version of pysam you should be able to reproduce the test fails locally. I've managed to resolve all the errors relating to the sam header on the {TS}-rebuild_tests branch but I still getting an unrelated error relating to the stats. When I've resolved this, I'll merge to master and you should be able to get your branch to pass.

TomSmithCGAT:
Right, tests are passing on the {TS}-rebuild_tests branch now. If you could review the pull request I'll merge into master and then we can merge master into this branch to fix this branch

TomSmithCGAT:
The tests pass when I re-ran them (i.e once the {TS}-rebuild_test has been merged into master) so this is ready to merge in master

PR #106, opened 2017-04-27
{ts} rebuild tests

TomSmithCGAT:


TomSmithCGAT:
- Updates all the sam file test references so that they contain a header
- Simplifies the stats aggregation which was causing errors in travis (presumably due to differences in pandas version but I never managed to reproduce this locally)
- Corrects a bug in the _per_umi_per_position.tsv output where the counts index excluded values observed only after deduplication

PR #122, opened 2017-05-09
{ts} add sort option

TomSmithCGAT:
This branch adds the `--sort-output` option for _dedup_ and _group_ and hardcodes the output from _count_ to be in feature (e.g gene) sorted order. 

The sort order for _count_ is enforced by only yielding umi counts per feature once all reads have been parsed. This could potentially take a considerable amount of memory. As a quick check, I tested out on a file with 3.4M reads, of which 1.7M are assigned to a gene and there are 0.1M true UMIs. The total memory requirement is ~130Mb which is OK but it does suggest this approach may not be suitable for very deep sequencing where many more UMIs are observed.


IanSudbery:
hmmm..... in the absolute worst case for a 8nt UMI we could have 20,000 genes with 65k UMIs. If each slot stores a 32bit int, and the key is an 8 byte string that means that gives a lower bound on the memory used in the worst case of 6.5GB .

We could go back to doing things per chrom I suppose.

IanSudbery:
Two things to ponder ... 
1) should we respect the OSes TMPDIR environment variable. 
2) Perhaps sorting should be default with the option to disable it if you want to avoid the temp file usage
3) In that case it should be disable (I would probably disallow anyway) if output is to stdout - surely using stdout suggests you want to save disk space.

TomSmithCGAT:
OK. I'll revert back to a per chrom approach for _count_ and sort from temp. Chrom 1 contains ~ 10% of all genes so this should reduce the memory requirement about 10-fold. If this isn't good enough, we can implement a heuristic solution at a later date - I'm thinking yielding genes when the input read being parsed is > 100Kbp after the last read for a gene and check to see that we don't 're-see' any yielded genes.

TomSmithCGAT:
1. I set it to the current directory for testing and forgot to finalise the tempdir location. Thanks!
2. I think this is perhaps sensible. Unless I can come up with a good reason (aside from the slight hassle re-making all the test files!) I'll do this.
3. I guess there could be applications where you want to pipe the output and need it sorted. I don't see any reason not disallow sorted output to stdout if we're allowing the user to disable anyway?

TomSmithCGAT:
Points 1 & 2 have been addressed. _count_ now yields after every contig, writes out a temporary file and sorts to stdout.

TomSmithCGAT:
I've merged this branch into my {TS}-support_cell_barcodes branch as I needed some of the same code so I'll close this pull request and open another for the later branch when it's ready

Issue #121, opened 2017-05-09
Add option to count using flatfile input


TomSmithCGAT:
Read counting tools such as featureCounts can output a flatfile containing the read ids and gene assignment. It would be good to include an option to input this flatfile rather than a BAM into umi_tools count. See comments by @MarinusVL in #44 for motivation

@IanSudbery - As I see it, all we need to do is make a simpler generator to be used in place of umi_methods.get_gene_count.

TomSmithCGAT:
Will be part of v0.5 release

PR #125, opened 2017-05-10
{ts} count from flatfile

TomSmithCGAT:
This introduces a _count_tab_ command which can be used to perform gene-level counting directly from the output of e.g featureCounts. 

The input format is hardcoded to be two columns, tab separated with read_id in column1 and assigned gene in column2, in gene order, with all non-assigned reads excluded. The command works from the stdin so the user can generate suitable input with command line tools if required and pipe to `umi_tools count_tab`.

This has been requested previously (see #44, #121) and allows us to better support scRNA-Seq analyses.


TomSmithCGAT:
This branch would benefit from some of the changes on the {TS}-Support_cell_barcodes branch. I'll close this oull request and merge in the {TS}-Support_cell_barcodes branch before putting in a new pull request.

Issue #111, opened 2017-05-05
umi_tools cellbarcode


TomSmithCGAT:
* This tool would take a BAM with a cell barcode and add a tag for cells which all came from the same cell

* Compute the tags using any of the UMI methods or the knee method

EDIT: From the discussion below these are the functionalities we need to add with UMI-tools _extract_. I suggest we do these sequentially (I'm adding this here to make the checklist visible in the issue description) :

- [x]  Extract cell barcodes. This should allow extraction from any encoding pattern. 
- [x]  Identify the 'true" cell barcodes by 'knee' method or network-based method.
- [x]  Error correct cell barcodes 

EDIT END.

TomSmithCGAT:
Text above from @IanSudbery's project note. 

I think there's an argument for performing the cell barcode grouping prior to alignment since all we need is the counts per cell barcode. The steps would be:

1. Parse the input fastqs and extract the cell barcode. Count reads per cell barcode
2. Identify the number of "true" cell barcodes (knee method or network-based method) -> barcode whitelist
3. Re-parse the input fastq, extract the UMI and cell barcode and write out all reads in the whitelist. We could also error-correct cell barcodes here. If knee method, correct barcodes within 1 edit distance of a single "true" barcode. If network-based method, correct all barcodes within the same group.

IanSudbery:
I suppose I was thinking that the pipeline would be:

1. Parse input fastqs and extract both cell bc and UMIs, add to read name
2. Align.
3. Process aligned file to convert the cell barcode stored in the name to a cell barcode tag, doing any grouping or correcting. 

I hadn't twigged that statistics on the barcodes could be output by extract and the grouping done at that point meaning we would only need two passes over the file - one to extract and group and one to label. My original pipeline would have had one to extract, one pass to do the group and correcting and one pass to label, where the final two passes would be done in a single tool call. 

Your way probably makes more sense.

TomSmithCGAT:
Ok so there are functionalities we need to add with UMI-tools _extract_. I suggest we do these sequentially:

- [ ]  Extract cell barcodes. This should allow extraction from any encoding pattern. 
- [ ]  Identify the 'true" cell barcodes by 'knee' method or network-based method.
- [ ]  Error correct cell barcodes

TomSmithCGAT:
[...]

Issue #112, opened 2017-05-05
Add ability to consider cell barcode with count command


TomSmithCGAT:

TomSmithCGAT:
@IanSudbery. Could you clarify exactly where we want to consider the cell barcode and for what purpose? E.g do we want to support cell barcodes for both _dedup_ and _count_? Are we going to output error-corrected cell barcodes?

IanSudbery:
I think it would be easy enough to support in both dedup and count.

TomSmithCGAT:
I'm moved this back to 'To Do' since we haven't actually finished off the handling of cell barcodes. The last step is to support cell barcodes within _group_, _count_ and _count_tab_

PR #128, opened 2017-05-19
{ts} extract cell barcodes

TomSmithCGAT:
Hi @IanSudbery. This covers the extraction of cell barcodes, filtering against a whitelist, automatic whitelist creation via 'knee' and 'directional' methods etc as discussed in #111.


TomSmithCGAT:
It's not clear why the the scipy installation failed with f66b054 but forcing pip install scipy prior to UMI-tools installation in travis resolves the problem.

TomSmithCGAT:
The whitelist detection method is now hardcoded to "knee" with the option to error correct barcodes within an edit distance threshold. I think we should also consider adding an option to filter out whitelisted barcodes within an edit distance threshold of another whitelisted barcode with higher counts where the whitelist is generated using the knee method - see #138 . I'll leave this for another branch.

@IanSudbery I'm going to merge this into master shortly so I can continue with supporting cell barcodes in _group_, _count_ etc - #112 .

Issue #131, opened 2017-06-02
Question about group / dedup --per-gene


Simon-Coetzee:
When I run the following command (or a similar dedup command):

`umi_tools group -I test.bam --per-gene --gene-transcript-map='gene2rapmap.tsv' --edit-distance-threshold=2 --group-out=test.bam.grouped.tsv --output-bam -S test.group.2.bam -L test.group.2.log --`

dedup command looks like this:
`umi_tools dedup -I test.bam --per-gene --gene-transcript-map='gene2rapmap.tsv' --edit-distance-threshold=2 -S test.dedup.2.bam -L test.dedup.2.log --`

where gene2rapmap.tsv looks like this:
```
ENSG00000223972.5       ENST00000456328.2|ENSG00000223972.5|OTTHUMG00000000961.2|OTTHUMT00000362751.1|DDX11L1-002|DDX11L1|1657|processed_transcript|
ENSG00000223972.5       ENST00000450305.2|ENSG00000223972.5|OTTHUMG00000000961.2|OTTHUMT00000002844.2|DDX11L1-001|DDX11L1|632|transcribed_unprocessed_pseudogene|
ENSG00000227232.5       ENST00000488147.1|ENSG00000227232.5|OTTHUMG00000000958.1|OTTHUMT00000002839.1|WASH7P-001|WASH7P|1351|unprocessed_pseudogene|
ENSG00000278267.1       ENST00000619216.1|ENSG00000278267.1|-|-|MIR6859-1-201|MIR6859-1|68|miRNA|
ENSG00000243485.5       ENST00000473358.1|ENSG00000243485.5|OTTHUMG00000000959.2|OTTHUMT00000002840.1|MIR1302-2HG-001|MIR1302-2HG|712|lincRNA|
ENSG00000243485.5       ENST00000469289.1|ENSG00000243485.5|OTTHUMG00000000959.2|OTTHUMT00000002841.2|MIR1302-2HG-002|MIR1302-2HG|535|lincRNA|
ENSG00000284332.1       ENST00000607096.1|ENSG00000284332.1|-|-|MIR1302-2-201|MIR1302-2|138|miRNA|
ENSG00000237613.2       ENST00000417324.1|ENSG00000237613.2|OTTHUMG00000000960.1|OTTHUMT00000002842.1|FAM138A-001|FAM138A|1187|lincRNA|
[...]

Issue #123, opened 2017-05-09
Remove "per-gene" options for group and dedup?


TomSmithCGAT:
We now support `-per-gene` with _group_, _count_ and _dedup_ commands. For _group_ and _dedup_, this method is problematic when used with `--gene-tag` since the BAM file may not be  sorted in gene-tag order due to overlapping genes. For the _count_ method, we can get around this by storing umi counts per gene until we reach the end of a chromosome since the counts take much less memory than storing the reads. 

Since the `--per-gene` will only be used in the context of read counting (as far as we're aware), should we remove the `--per-gene` options from _group_ and _dedup_? This would also help make clearer the different functions of the three commands.

We would obviously need to make users aware of this change and perhaps check for use of the `--per-gene` option when using _dedup_ and _group_ to explain that this is only available in _count_ now.

TomSmithCGAT:
We're going to remove the option to run _group_ and _dedup_ with `--per-gene --gene-tag`, thus we wont support alignment to a genome and -per-gene deduplication or grouping. It will still be possible to count per-gene using a gene-tag.

The use of '--per-gene --gene-tag' with _dedup_ or _group_ will raise an error and suggest the user can use _count_ if they wish to obtain per-gene UMI counts

TomSmithCGAT:
During the development of the count command and all the single cell changes, the options for --per-gene analyses and how these are handled have become a little messy. I'm going to recode so all commands use the same logic internally regarding per-gene grouping/deduplication/counting. Thus, for the following alignments, the following options will be required.

- Alignment to a genome + gene assignment with e.g featureCounts: Use the  `--per-gene`  and `--gene-tag` options. `get_bundles` will yield the entire reads_dict whenever the contig changes. This ensures all reads for each gene on the contig have been parsed

- Alignment to a 'super-transcriptome': Use the `--per-gene` and `--per-contig` options. `get_bundles` will yield the entire reads_dict whenever the contig changes since each contig here is a single gene. The same settings could theoretically be used to group/dedup/count on a per transcript isoform level but we will make clear this is not recommended

- Alignment to a transcriptome: Use the `--per-gene`, `--per-contig` and  `-gene-transcript-map` options. `get_bundles` will only yield if all transcripts (contigs) for the gene have been observed. This ensures all reads for all transcripts for the gene have been parsed.

This set up will also enable the user to perform per-position counting if required. In fact, this will be the default for _count_ with the `--per-gene` option triggering per-gene counting.

[...]

PR #139, opened 2017-06-20
{ts} support cell barcodes

TomSmithCGAT:
This branch includes includes the following updates:

- Output of _dedup_/_count_ is now sorted by default (from the {TS}-Add_sort_order branch)
- _count_ now includes a `--per-cell` option to count reads per gene per cell
- New umi extraction method, "umis", for reads where extraction was done with [umis](https://github.com/vals/umis)
- Adds whole-contig method to _group_, as per _dedup_
- Improved documentation


Simon-Coetzee:
I think that there may be a typo here:

https://github.com/CGATOxford/UMI-tools/blob/%7BTS%7D-support_cell_barcodes/umi_tools/count.py#L316

where it says `gene_tag=options.gene_tag` i believe it should just be `gene_tag=gene_tag` because it's redefined above (and `options.gene_tag` is `none` if it's not defined because you're using `gene_transcript_map`)

While with that change, i am able to proceed, I then get stuck on line 330 with this error
```
Traceback (most recent call last):
  File "/home/coetzeesg/miniconda3/envs/test_umi_count/bin/umi_tools", line 11, in <module>
    load_entry_point('umi-tools==0.4.4', 'console_scripts', 'umi_tools')(
  File "/home/coetzeesg/miniconda3/envs/test_umi_count/lib/python2.7/site-packages/umi_tools-0.4.4-py2.7.egg/umi_tools/umi_tools.py", line 46, in main
    module.main(sys.argv)
  File "/home/coetzeesg/miniconda3/envs/test_umi_count/lib/python2.7/site-packages/umi_tools-0.4.4-py2.7.egg/umi_tools/count.py", line 333, in main
    threshold=options.threshold)
  File "/home/coetzeesg/miniconda3/envs/test_umi_count/lib/python2.7/site-packages/umi_tools-0.4.4-py2.7.egg/umi_tools/network.py", line 400, in __call__
    assert max(len_umis) == min(len_umis), (
ValueError: max() arg is an empty sequence
```
given the command
`umi_tools count -I test.bam --per-cell --gene-transcript-map=gene2transcript.tsv -L test.log -S test_count.tsv --wide-format-cell-counts`
same error if i leave off `--wide-format-cell-counts`
and i can't quite follow why I get that error.

@TomSmithCGAT Thanks for this branch. It's exactly what I need.

Simon-Coetzee:
By the way, I think that was the reason I was getting blank gene gene names at the end of #131

TomSmithCGAT:
Thanks for picking this up @Simon-Coetzee. 

The error message indicates the the UMIs extracted from the read are not all the same length. I'm assuming your UMIs are actually all the same length and this is actually an error relating to how to the UMIs are extracted from the read.

This branch also introduces the `--extract-umi-method=umis` option which will handle read names where the UMIs have been attached using vals umis package. Could you try again using this option.

Simon-Coetzee:
I'm not certain that that is the problem, I have used your extract method, because I like the filtering for valid barcodes based on the knee method you have. My reads names look like this:
```
NS500624:101:H2WWHBGX2:2:12104:13927:12357_TCGGTAAAGGGTATCG_ATCGGTGTAG
NS500624:101:H2WWHBGX2:4:12611:22668:13471_TCGGTAAAGGGTATCG_ATCGGTTTAG
NS500624:101:H2WWHBGX2:2:21102:19002:4478_CACAAACTCGTTTATC_GCCACAGGAT
NS500624:101:H2WWHBGX2:3:11608:19732:3472_TCGGTAAAGGGTATCG_ATCGGTTTAG
NS500624:101:H2WWHBGX2:1:12210:18502:2899_TCGGTAAAGGGTATCG_ATCGGTTTAG
NS500624:101:H2WWHBGX2:3:11411:10450:5402_TCGGTAAAGGGTATCG_ATCGGTTTAG
NS500624:101:H2WWHBGX2:1:23201:5661:15259_CGCTATCCACATGTGT_GACTACGGCC
NS500624:101:H2WWHBGX2:2:11208:5535:3022_TCGGTAAAGGGTATCG_ATCGGTTTAG
NS500624:101:H2WWHBGX2:1:21201:13270:19241_TCGGTAAAGGGTATCG_ATCGGTTTAG
NS500624:101:H2WWHBGX2:4:11501:26307:11543_CGCTATCCACATGTGT_GACTACGGCC
```
If i parse the length of the trailing UMI's they're all exactly length 10.
The error in the assert is not that they don't match, but rather that `max() arg is an empty sequence`

So I think, but don't know for sure, but somewhere in count.py lines 310-323 the umi's object is being created strangely for my file (maybe it's malformed in some way that i'm not aware of).

if i stick a print statement in line 328, thusly:
```
processor = network.UMIClusterer("directional")
print "counts = %s ; cell = %s ; gene = %s" % (counts, cell, gene)

# group the umis
groups = processor(
        umis,
        counts,
        threshold=options.threshold)
```

I get this readout
```
counts = {'CGGGCGAGCC': 6, 'CGGGCGAGCT': 1, 'CGGGCGAGCG': 1} ; cell = GTTAAGCTCCCTCTTT ; gene = ENSG00000237491.8
counts = {'GTCCGGTGGT': 1} ; cell = ACATGGTCAATCTGCA ; gene = ENSG00000237491.8
counts = {'GTCCGGTGGT': 1} ; cell = ACATGGTCAATCTGCA ; gene = ENSG00000237491.8
counts = {'CACTAATGCC': 1, 'CTGTAATGCG': 4} ; cell = TGGACGCAGAGTCGGT ; gene = ENSG00000237491.8
counts = {'CGGGCGAGCC': 1} ; cell = GTTAAGCTCCCTCTTT ; gene = ENSG00000237491.8
counts = {'ACCTCCCGGC': 1, 'ACCTCCCGGA': 7} ; cell = TTGACTTCAGGACCCT ; gene = ENSG00000237491.8
counts = {'ACCTCCCGGC': 1, 'ACCTCCCGGA': 7} ; cell = TTGACTTCAGGACCCT ; gene = ENSG00000237491.8
counts = {'CGGGCGAGCC': 1} ; cell = GTTAAGCTCCCTCTTT ; gene = ENSG00000237491.8
counts = {'CCCAGGTTGA': 1} ; cell = CTAGTGATCGAGAACG ; gene = ENSG00000237491.8
counts = {'TTCGTTCATT': 1, 'TGAATCTCTA': 1} ; cell = GCACTCTCACAGGAGT ; gene = ENSG00000237491.8
counts = {'CGGGCGAGCC': 1} ; cell = GTTAAGCTCCCTCTTT ; gene = ENSG00000237491.8
counts = {'ACCTCCCGGC': 1, 'ACCTCCCGGA': 8} ; cell = TTGACTTCAGGACCCT ; gene = ENSG00000237491.8
counts = {'GATCCGCTCG': 1, 'GCTCCGCTCC': 1, 'GCTCCGCTCG': 3} ; cell = ATCACGAGTTGGTAAA ; gene = ENSG00000237491.8
counts = {'ACCTCCCGGC': 1, 'ACCTCCCGGA': 8} ; cell = TTGACTTCAGGACCCT ; gene = ENSG00000237491.8
counts = {'CGGGCGAGCC': 1} ; cell = GTTAAGCTCCCTCTTT ; gene = ENSG00000237491.8
counts = {'GTCCGGTGGT': 1} ; cell = ACATGGTCAATCTGCA ; gene = ENSG00000237491.8
counts = {'CGGGCGAGCG': 1, 'CGGGCGAGCC': 19, 'CGGGCGAGCT': 1, 'CGGGCGAGGC': 1, 'CGGGCGATCC': 1} ; cell = GTTAAGCTCCCTCTTT ; gene = ENSG00000237491.8
counts = {'ACCTAAGAAT': 2} ; cell = CCTAAAGAGTGGACGT ; gene = ENSG00000237491.8
counts = {'CCTTGTGATG': 1} ; cell = AGGTCCGAGCCACGCT ; gene = ENSG00000162571.13
counts = {} ; cell = AGGTCCGAGCCACGCT ; gene = ENSG00000237491.8
```

Where it fails on the last one, because there are no UMI counts for that cell/gene combination?

Thanks for your help on this

TomSmithCGAT:
Yes, sorry, you're right. I skimmed the error message and failed to actually read the ValueError message and just assumed it was the assert which failed :man_facepalming:

I'll have a look at what might be going wrong. This will be quicker with a test case which fails. Would you be able to share the BAM which is failing, subsetted to just ENSG00000237491.8. You can email me at tss38@cam.ac.uk if the subsetted BAM is small enough.

P.s glad you like the new cell barcode extraction :grinning:

TomSmithCGAT:
Ah it looks pretty simple. There was a bug in the  `get_gene_count` generator such that when the contig changes, it yields the wrong read counts. Any remaining counts are yielded (correctly) at the end of the file. This wasn't picked up in the tests since the test input has just a single contig. 

Your output above still looks strange though. The generator should only yield for each gene once so you shouldn't see repeated lines with the same gene_id. The only way I can see that you would yield for the same gene more than once is when you have reads on different contigs assigned to the same gene! Do you have multimapping reads perhaps? Again, probably easiest for me to take a look at your subsetted BAM to work out what's going on here.

TomSmithCGAT:
Thanks for sharing your BAMs

The get_gene_count generator yields whenever the contig changes in order to keep prevent the counts_dict object (tallies the number of times each genes:cell:umi combination has been observed) getting to large. This is fine when the reads are aligned to a genome, since the BAM is sorted by position so whenever the contig (chromosome) changes, we can yield all the tallies as the same genes will not be observed again. However, we the reads are aligned to a transcriptome, when the contig (transcript) changes, this does not mean all reads for a gene have been observed. I've fixed this so the generator only yields when all transcripts for a given gene have been observed.

Would you mind checking whether the latest commit works in your hands?

TomSmithCGAT:
I think this branch is finally ready to go (assuming the tests pass!). Would you like to give it a once over before I merge? Please see #123 for a more detailed explanation of the most recent updates

Simon-Coetzee:
Okay, running this version on the the bam I sent you before: `AGGTCCGAGCCACGCT_cell.bam`
gets me an error run like this with python3:

`umi_tools count -I AGGTCCGAGCCACGCT_cell.bam --per-contig --gene-transcript-map=small.gene2trans.txt -L test.log -S test.count.txt`

here's the error:
```
Traceback (most recent call last):
  File "/home/coetzeesg/miniconda3/envs/umi_count_3/bin/umi_tools", line 11, in <module>
    load_entry_point('umi-tools==0.4.4', 'console_scripts', 'umi_tools')()
  File "/home/coetzeesg/miniconda3/envs/umi_count_3/lib/python3.6/site-packages/umi_tools-0.4.4-py3.6.egg/umi_tools/umi_tools.py", line 47, in main
    module.main(sys.argv)
  File "/home/coetzeesg/miniconda3/envs/umi_count_3/lib/python3.6/site-packages/umi_tools-0.4.4-py3.6.egg/umi_tools/count.py", line 163, in main
    tmpfile.write("%s\n" % ":".join((gene, str(gene_count))))
TypeError: sequence item 0: expected str instance, int found
```

If i change `umi_methods.py` thusly:
```
diff --git a/umi_tools/umi_methods.py b/umi_tools/umi_methods.py
index 619feaa..0b86506 100644
--- a/umi_tools/umi_methods.py
+++ b/umi_tools/umi_methods.py
@@ -1170,7 +1170,7 @@ class get_bundles:
             if self.options.per_gene:

                 if self.options.per_contig:
-                    pos = read.tid
+                    pos = read.reference_name

                 elif self.options.gene_tag:
                     pos = read.get_tag(self.options.gene_tag)
@@ -1205,7 +1205,7 @@ class get_bundles:
                 start, pos, is_spliced = get_read_position(
                     read, self.options.soft_clip_threshold)

-                current_chr = read.tid
+                current_chr = read.reference_name
                 do_output, out_keys = self.check_output(pos)

                 if do_output:
```

And then rerun with the same invocation of `count` as above, I get this error:

```
Traceback (most recent call last):
  File "/home/coetzeesg/miniconda3/envs/umi_count_3/bin/umi_tools", line 11, in <module>
    load_entry_point('umi-tools==0.4.4', 'console_scripts', 'umi_tools')()
  File "/home/coetzeesg/miniconda3/envs/umi_count_3/lib/python3.6/site-packages/umi_tools-0.4.4-py3.6.egg/umi_tools/umi_tools.py", line 47, in main
    module.main(sys.argv)
  File "/home/coetzeesg/miniconda3/envs/umi_count_3/lib/python3.6/site-packages/umi_tools-0.4.4-py3.6.egg/umi_tools/count.py", line 136, in main
    for bundle, key, status in bundle_iterator(inreads):
  File "/home/coetzeesg/miniconda3/envs/umi_count_3/lib/python3.6/site-packages/umi_tools-0.4.4-py3.6.egg/umi_tools/umi_methods.py", line 1125, in __call__
    for read in inreads:
  File "/home/coetzeesg/miniconda3/envs/umi_count_3/lib/python3.6/site-packages/umi_tools-0.4.4-py3.6.egg/umi_tools/umi_methods.py", line 925, in metafetcher
    for metacontig in metacontig2contig:
RuntimeError: dictionary changed size during iteration
```

I think it's happening when when one gene/metacontig is complete, and it's switching over to the next metacontig, but I'm not sure.

I've attached my gene-transcript-map here: [small.gene2trans.txt](https://github.com/CGATOxford/UMI-tools/files/1140497/small.gene2trans.txt) in case that's a problem too.

Documentation and everything else looks good! Thanks

Simon-Coetzee:
Okay I took a look at how the dictionary for metacontig2contig is changing in `metafetcher`.
For some reason, right before it dies it changes the metacontig2contig by adding these three elements to it:
```
{'': set(), 'ENST00000471248.1': set(), 'ENST00000466430.5': set()}
```

I don't know why.

TomSmithCGAT:
Hi @Simon-Coetzee. I can confirm I get the same error message with your input. The code is passing the functionality tests so I'm hopeful this is a minor bug relating to your specific input. I'll take a look now. Thanks for all your continued involvement!

TomSmithCGAT:
@Simon-Coetzee, could you try running `umi_tools count` with the latest updates to this branch? I think everything should be working OK now. 

There were some bugs in the updated code specifically relating to counting when using the --gene-transcript-map option. This option isn't currently covered by tests hence the bugs slipping through. Would you mind if I used a subset of your input as a test file?

Simon-Coetzee:
Looks good, I've run it successfully on some small files (like that bam I set you - which you are free to use if you'd like for testing purposes).

I had to make one small change due to a typo:

```
diff --git a/umi_tools/umi_methods.py b/umi_tools/umi_methods.py
index 55571e4..ed1c7f2 100644
--- a/umi_tools/umi_methods.py
+++ b/umi_tools/umi_methods.py
@@ -994,7 +994,7 @@ class get_bundles:
         self.metacontig_contig = metacontig_contig
 
         self.contig_metacontig = {}
-        if self.contig_metacontig:
+        if self.metacontig_contig:
             for metacontig in metacontig_contig:
                 for contig in metacontig_contig[metacontig]:
                     self.contig_metacontig[contig] = metacontig
```

But that's it. After I finish a full sample I'll let you know that it worked on a full sized dataset too.

Thanks!

Simon-Coetzee:
I do have a question, about getting gene names in the tags of the bams like you seem to prefer (it makes good sense to me too).  Given the output of STAR or Rapmap, or anything like that mapped to the genome, what tools do you use to add gene names to them?

Thanks

TomSmithCGAT:
Thanks for spotting the typo! 

We don't really have a preference for gene names in the tag or as field 3 in the BAM. I only recently realised that STAR could align to the genome and output a separate reads with the reads aligned to the transcriptome. This seems like a really neat solution to me. If you did want to get gene assignment in a tag then you can run e.g featureCounts/HTSeq on the genome BAM. My understanding from a few forum posts is that the gene assignments are essentially identical to running STAR as above, but the STAR approach is quicker as it's all done in one go.

README.rst
```diff
@@ -5,19 +5,32 @@ UMI-tools was published in `Genome Research <http://genome.cshlp.org/content/ear
...
-This repository contains tools for dealing with Unique Molecular Identifiers (UMIs)/Random Molecular Tags (RMTs). Currently there are three tools:
+This repository contains tools for dealing with Unique Molecular
+Identifiers (UMIs)/Random Molecular Tags (RMTs). Currently there are
+five tools
 
 * extract:
    **Flexible removal of UMI sequences from fastq reads.**
       UMIs are removed and appended to the read name. Any other barcode, for example a library barcode, is left on the read.  
 
+* group:
```
IanSudbery: whitelist?

umi_tools/whitelist.py
```diff
@@ -0,0 +1,456 @@
...
+
+try:
+    import umi_tools.Utilities as U
+except ImportError:
+    import Utilities as U
+
+try:
+    import umi_tools.network as network
+except ImportError:
+    import netwxoork
```
TomSmithCGAT: Thanks!

umi_tools/whitelist.py
```diff
@@ -0,0 +1,456 @@
...
+
+    (options, args) = U.Start(parser, argv=argv,
+                              add_group_dedup_options=False,
+                              add_sam_options=False)
+
+    if not options.pattern and not options.pattern2:
+        if not options.read2_in:
+            U.error("Must supply --bc-pattern for single-end")
+        else:
+            U.error("Must supply --bc-pattern and/or --bc-pattern "
```
TomSmithCGAT: Thanks!

umi_tools/whitelist.py
```diff
@@ -0,0 +1,456 @@
...
+        if options.extract_method == "string":
+            U.error("barcode pattern(s) do not include any cell bases "
+                    "(marked with 'Cs') %s, %s" % (
+                        options.pattern, options.pattern2))
+        elif options.extract_method == "regex":
+            U.error("barcode regex(es) do not include any cell groups "
+                    "(starting with 'cell_') %s, %s" (
+                        options.pattern, options.pattern2))
+
+    if options.stdin == sys.stdin:
```
TomSmithCGAT: You're right, this is uneccessary since `options.stdin` is handled by U.start such that it will always be a file object so lines 356-359 can be replaced by `umi_methods.fastqIterate(options.stdin)` . This was copied from umi_tools.extract so I'll remedy there too.

umi_tools/whitelist.py
```diff
@@ -0,0 +1,456 @@
...
+                else:
+                    cell_barcode_counts[cell] += 1
+                n_cell_barcodes += 1
+
+            if options.subset_reads:
+                if n_cell_barcodes > options.subset_reads:
+                    break
+    else:
+        read2s = umi_methods.fastqIterate(U.openFile(options.read2_in))
+        for read1, read2 in izip(read1s, read2s):
```
TomSmithCGAT: I'll merge master into this branch to bring in your fastq parsing commit (9d8c1eae6854569bf32bf2d0ef57cfb03b0ccead)

umi_tools/whitelist.py
```diff
@@ -0,0 +1,456 @@
...
+
+            if options.subset_reads:
+                if n_reads > options.subset_reads:
+                    break
+
+    if options.method == "umis":
+        for cell in cell_barcode_umis:
+            cell_barcode_counts[cell] = len(cell_barcode_umis[cell])
+
+    if options.cell_number and options.cell_number > len(cell_barcode_counts):
```
TomSmithCGAT: Yes, this means the user has specified that there are , e.g 1000 cells in the sample, but <1000 unique cell barcodes were observed.

umi_tools/whitelist.py
```diff
@@ -0,0 +1,456 @@
...
+                options.cell_number, len(cell_barcode_counts)))
+
+    cell_whitelist, true_to_false_map = umi_methods.getCellWhitelist(
+        cell_barcode_counts,
+        options.cell_number,
+        options.error_correct_threshold,
+        options.plot_prefix)
+
+    # re-make the reads1s iterator
+    read1s = umi_methods.fastqIterate(U.openFile(options.stdin.name))
```
TomSmithCGAT: Thanks for spotting. This bug is left-over from when the code was used to both generate the whitelist and then filter the fastq so the iterator was remade. This would of course cause a fail when `options.stdin`==`sys.stdin`!

umi_tools/whitelist.py
TomSmithCGAT: I was being silly. Just realised `joinedFastqIterate` is only used in `extract` and required to handle cases where the user has already filtered the reads for some reason and the reads are not strictly paired, as per the output from `alevin`. For `whitelist` we can reasonably demand that the user has not performed any pre-filtering as this would invalidate the assumptions used to define the whitelist so we can stick with the above `izip(read1s, read2s)`

README.rst
TomSmithCGAT: Good point. `README.rst` should also describe `count` and `count_tab` since these commands are also now available.

umi_tools/whitelist.py
IanSudbery: Perhaps sampling means that RNA from every cell was not sequencing?

TomSmithCGAT: Fair enough, I'll update the error message to reflect this.

PR #141, opened 2017-07-03
Update network.py

willchen108:
No need to compare all umi1 umi2 in umis. It's redundant. only need to compare unique umis in counts.keys(). Won't be an issue for thousands of umis. but it's a pain in my case, which has hundreds of thousands of umis. Take up a lot memories and time consuming

### DIFF
--- a/umi_tools/network.py
+++ b/umi_tools/network.py
@@ -233,7 +233,7 @@ class UMIClusterer:
         ''' identify all umis within hamming distance threshold'''
 
         adj_list = {umi: [] for umi in umis}
-        for umi1, umi2 in itertools.combinations(umis, 2):
+        for umi1, umi2 in itertools.combinations(counts.keys(), 2):
             if edit_distance(umi1, umi2) <= threshold:
                 adj_list[umi1].append(umi2)
                 adj_list[umi2].append(umi1)
@@ -245,7 +245,7 @@ class UMIClusterer:
         and where the counts of the first umi is > (2 * second umi counts)-1'''
 
         adj_list = {umi: [] for umi in umis}
-        for umi1, umi2 in itertools.combinations(umis, 2):
+        for umi1, umi2 in itertools.combinations(counts.keys(), 2):
             if edit_distance(umi1, umi2) <= threshold:
                 if counts[umi1] >= (counts[umi2]*2)-1:
                     adj_list[umi1].append(umi2)


TomSmithCGAT:
The comparison is already already between only the UMI in counts.keys since counts is created from umis. See network.py lines 436-437

```python
umis = bundle.keys()
counts = {umi: bundle[umi]["count"] for umi in umis}
```

As it happens, I was working yesterday on estimating the memory usage for a given number of UMIs and UMI length. From version 0.5 onwards, a warning will be raised when the memory usage is expected to be very high. As a rough guide, 10,000 unique UMIs at a given position will require ~2GB, 100,000 ~17Gb. We can look at reducing this requirement by building and resolving each individual network at a position individually rather than storing all the edges for all networks at the same time as we currently do. This will increase the run time however, and wont help when the number of UMIs approaches the total number of possible UMIs (e.g 4^UMI length).

willchen108:
Thanks Tom. Mine case has more than 100,000 UMIs and it takes more than 100G memory which sounds really crazy. I also did some test. The memory usage is always hight than yours. My UMI is 10bp long. I realize the breadth_first_search also is also really time consuming, is it possible to make it cython as well(or make it parallel)? As the graph size increase, search for connected nodes will be really time consuming.

TomSmithCGAT:
Hi @willchen108. At 100,000 UMIs, you've covered 9.5% of all possible UMIs - assuming you mean unique UMIs? This is too much for any of assumptions about the 'uniqueness' of the UMIs to hold. I would therefore recommend repeating the experiment with a longer UMI (>=14bp) to accommodate such deep sequencing. If this is not possible,  I would suggest using the 'unique' method in UMI-tools to analyse this data, with the obvious implications that this will not deal with errors.

willchen108:
Hi Tom, thanks. That's a good suggestion. I already change the UMI to be 16bp and hopefully it works better.
Thanks

k3yavi:
Coming to think of it, the difference in the memory usage between the experiments run by @willchen108 and @TomSmithCGAT could be because it would depend, if I am right, also on the size of the connected components in the graph along with, and in addition to the total number of unique UMI. Moreover, parallelization would not really help if the memory is taking the hit because of the former case which probably looks like the case with the data of @willchen108.
Also, if I may ask @willchen108 after reading your last comment I am wondering is it really this easy to change the size of the UMI biologically?

TomSmithCGAT:
Hi @k3yavi. You're correct. Strictly speaking the memory usage is dependent upon the size of the connected components in the graph, which is in turn dependent on the number of UMIs and the UMI length. At the moment we perform the following steps for each position given the set of UMIs (and their frequency):

1. All-vs-all edit-distance comparison (+ frequency with directional method) to identify all edges between all nodes (UMIs) in all connected components --> `adj_list` dictionary
2. Identify all the connected components from `adj_list`
3. Resolve each connected component in turn in a method-dependent manner

The major memory usage issue is likely to be the `adj_list`. I did some very naive simulations which we'd hoped to use to estimate the memory footprint of `adj_list`. However, Ian has shown that the estimates of memory usage from the simulations don't accurately reflect real data. In theory it should be possible to at least roughly estimate the memory usage and run time at each position given the number of UMIs and the UMI length and this is something we'd like t o return too at some point so that we can at least warn the user that the memory usage and/or run time estimated to be high.

As I said in the comment above, we could reduce the memory usage by identifying the edges for each connected component individually. For the directional method the steps would be:

Starting from the set of all UMIs identified at a given position, and their frequency:
1. Take the most abundant UMI and identify all edges
2. Repeat Step 1 for all new UMIs connected by an edge to search through the graph for all edges for the connected component
3. Once no more new UMIs are found, resolve the connected component and remove UMIs from set
4. Go back to step 1 until the set of UMIs is empty

This could potentially massively reduce the memory usage in situations where the number of UMIs is high (e.g 100,000) but the length is sufficiently long that the UMIs don't all just form one big horrible component. 

I've not got any experience with parallelisation in python. I guess for optimal run time we could use multiple threads for step 3 above but my guess is that the great majority of the run time will be steps 1 and 2. 

Maybe you have a more elegant solution?

TomSmithCGAT:
@willchen108 Glad to hear it. Good luck with the new experiments!

k3yavi:
Hi @TomSmithCGAT ,

I really like the ideas you proposed and it looks like you definitely have invested a lot of time into this. But, out of my curiosity, I'd really like to ask a couple of more questions since they can help solve the problem better:
(1) If I understood it right you are maintaining `adj_list` dictionary for each position i.e. |`adj_list`| would be the total number of unique genomic positions in the `Sam` file. One useful figure would be to plot, is the distribution of the lengths of the connected components along the whole `adj_list`. Basically, what I want to know is, what is blowing up the memory a few very big connected components or very large number of small components.
(2) I was wondering wouldn't increment in the size of the barcode increase the complexity of the problem? If I am right then this would increase the length of the connected component even more since the possible number of possible variations from the true UMI would increase or I am missing something very basic?

Also a couple of suggestions:
(1) I like the simplified process proposed by you for UMI clustering but IMO using parallelization at inter-position level would be much easier than using at intra-position level since each task would be independent.
(2) Also to speed up the process we can pre-index the edges i.e. for all UMI we can save all 1-edit distance UMI from it so that query would be faster but this becomes computationally infeasible with the relatively bigger size of the barcodes (up to 16 is feasible).

Hope it helps!

TomSmithCGAT:
Hi Avi, 

Very happy to keep discussing. We should probably move this over into an issue though so it doesn't get forgotten about!

Regarding the memory usage:
1) I may have slightly misunderstood what you meant here but for clarification, we don't maintain an `adj_list` for each position. We generate a new `adj_list` for each position. We could definitely have a look at logging some statistics on the size of the components and memory usage as we go though so we can interrogate this at the end of the run. I agree it would be very interesting to see how the number of UMIs and UMI length relate to the number and size of components and how these relate to the memory usage

2) You're correct that if everything is held the same, increasing the length of the UMI will slightly increase the number of error UMIs detected (and thus the length of the connected component) since the probability of having zero errors across the UMI increases. However, In the case we are discussing above, the number of UMIs sequenced is close to the total number of all UMIs possible (100,000 UMIs = ~10% of all 10bp UMIs). Thus each connected component will contain many hubs (true UMIs), increasing its complexity. I would assume that the complexity of the connected components would be reduced by increasing the length of the UMI to e.g 16bp since this will increase the total number of possible UMIs 4096-fold and thus allow the hubs to fall into separate components. We could test this explicitly by reverse engineering smaller UMIs from longer ones. For example, if we had some lovely gold standard 16bp UMI data with up to 100,000 unique UMIs at a given position, we could truncate the UMIs to 15,14,...,10bp and look at the complexity of the connected components for each UMI length. Combined with sub-sampling the reads this would give us a good idea of the relationship between the factors and the component complexity, memory usage and run-time.

Regarding parallelisation:
1) Inter-position parallelisation would indeed be a much more sensible approach. Indeed this is the only form of parallelisation we have considered. In fact I'm sure @IanSudbery  tried this but found the run-time improvement was surprisingly low. I'll dig out the issue where this was discussed.
2) We've discussed this as well but never tried implementing it. One thing to consider though is that for very long UMIs (>=16bp) the probability of getting error barcodes at an edit distance of 2 increases considerably so an edit distance threshold of 2 may be preferable. The exception being where the sequencing error rate is very low or the read depth is very high (Not sure what 'very' quantify-ably means in either sense here). In the later case this is because high read depth means you are more likely to also observe the intermediate 1-edit distance error barcode so the 2-error is also in the same connected component.

k3yavi:
Thanks for the clarification @TomSmithCGAT it does make sense now and yes I agree we should move this to a new issue.
However, now I am really excited to see the connected components' length distribution plot for the real data since it would help us understand the fundamental bottleneck with the problem. You can imagine if indeed there are multiple relatively smaller length components then parallelizing should help reduce the time and hopefully memory. But we might have to think a little more if it's the case that there are a few very big components which are creating the issue then parallelization can potentially help practically in expectation but theoretically the worst case would not be very different wrt memory.

PR #152, opened 2017-08-05
{ts} support cell barcodes

k3yavi:
Trivial mods for monitoring the progress of the extract method.
I believe if not present in other commands should/can be added.

### DIFF
--- a/umi_tools/extract.py
+++ b/umi_tools/extract.py
@@ -425,6 +425,10 @@ def main(argv=None):
             options.stdout.write(str(new_read) + "\n")
 
     else:
+        #variable for progress monitor
+        progCount = 0
+        displayMax = 500000
+       
         read2s = umi_methods.fastqIterate(U.openFile(options.read2_in))
 
         if options.read2_out:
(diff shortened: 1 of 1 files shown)


umi_tools/extract.py
```diff
@@ -425,6 +425,11 @@ def main(argv=None):
             options.stdout.write(str(new_read) + "\n")
 
     else:
+        #variable for progress monitor
+        progCount = 0
+        displayMax = 100000
+        print("Starting barcode extraction")
```
TomSmithCGAT: This would be better as a call to U.info() to send this to the logfile.

umi_tools/extract.py
```diff
@@ -437,6 +442,13 @@ def main(argv=None):
 
         for read1, read2 in umi_methods.joinedFastqIterate(
                 read1s, read2s, strict):
+            # incrementing count for monitoring progress
+            progCount += 1
+            # Update display in every 100kth iteration
+            if (progCount % displayMax == 0):
+                print("\r Parsed {} reads".format(progCount)),
```
TomSmithCGAT: Ditto above re. U.info()

umi_tools/extract.py
k3yavi: Yep, sounds good.
I was just trying to point it out as an example since there can be other such places where progress monitoring make sense, specially with 10x data where number of cells are easily above 1000 making library size even bigger.

PR #153, opened 2017-08-11
Update deduplicated_edit_distance.tsv's description

popucui:
From the context of Step5 Deduplication, in deduplicated_edit_distance.tsv file's third line, the positions where the average edit distance between UMIs is 1 should be '1167', not '11167'.


TomSmithCGAT:
Cheers!

PR #154, opened 2017-08-15
{ts} count from flatfile

TomSmithCGAT:
Resurrects  #125 

This introduces the `count_tab` command to count reads per gene using the flatfile output from e.g FeatureCounts/HTSeq. Currently `count_tab` does not support per-cell counting.


umi_tools/count_tab.py
```diff
@@ -0,0 +1,131 @@
...
+    parser = U.OptionParser(version="%prog version: $Id$",
+                            usage=globals()["__doc__"])
+
+    # add common options (-h/--help, ...) and parse command line
+    (options, args) = U.Start(parser, argv=argv, add_group_dedup_options=False)
+
+    if options.random_seed:
+        np.random.seed(options.random_seed)
+
+    if options.stdin != sys.stdin:
```
IanSudbery: This should not be neccessary. Experiment.py should do this for you. We only need this if we are handling a BAM

umi_tools/count_tab.py
```diff
@@ -0,0 +1,131 @@
...
+            infile,
+            umi_getter=umi_getter):
+
+        umis = counts.keys()
+
+        nInput += sum(counts.values())
+
+        # set up UMIClusterer functor with methods specific to
+        # specified options.method
+        processor = network.UMIClusterer(options.method)
```
IanSudbery: SHuoldn't need to build a new procesor each time

umi_tools/count_tab.py
```diff
@@ -0,0 +1,131 @@
...
+import sys
+import collections
+import re
+
+# required to make iteritems python2 and python3 compatible
+from builtins import dict
+
+from functools import partial
+
+import pysam
```
IanSudbery: Looks as if quite a lot of these imports are not neccesary. I don't see pysam used anywhere. Nor pandas, nor numpy.

umi_tools/umi_methods.py
```diff
@@ -1394,6 +1406,47 @@ def __call__(self, inreads):
                 yield self.reads_dict[p][k], k, "bundle"
 
 
+def get_gene_count_tab(infile,
+                       umi_getter=None):
+
+    ''' Yields the counts per umi for each gene
+
+    ignore_umi: don't include the umi in the dict key
```
IanSudbery: This and skip_regex do not appear to be arguments to `get_gene_count_tab`

umi_tools/count_tab.py
TomSmithCGAT: Yep, that's in the wrong place! Cheers

umi_tools/count_tab.py
TomSmithCGAT: Agreed lines 87-88 are unnecessary. Could you explain what the issue is with line 90?

umi_tools/count_tab.py
TomSmithCGAT: Yes, lazy copying...

umi_tools/umi_methods.py
TomSmithCGAT: Yes, `ignore_umi` is not a required method. `skip_regex` would be useful but not currently available. I've updated documentation to mark `skip_regex` as "TODO". Thanks!

umi_tools/count_tab.py
IanSudbery: If `options.stdin != sys.stdin`, then `options.stdin` should already be an open file handle pointing at the file specified by `-S` on the command line. 

image user runs `umi_tools count_tab -S feature_counts.tsv.gz`
Thus, with your code here on line 89 options.stdin is `<FileHandle: gzip <feature_counts.tsv.gz>>`. On line 90 you discover that it is not `stdin`, and so on line 91 you open the file `feature_counts.tsv.gz` again and set it to `infile`. 
Thus on line 95, `infile` contains what `options.stdin` contained on line 89, irrespective of what happens in 90-94, and the whole block could be replaced with a single line `infile=options.stdin`, or even `options.stdin` passed to `get_gene_count_tab` on line 104.
Bascially what

TomSmithCGAT: Ah yes, of course. I was forgetting how Experiment/Utilities handled the I/O options. Thanks!

PR #157, opened 2017-08-15
added import at start of umi_tools

Hoohm:
umi_tools crashes when running on a system without a graphical interface.

```  File "python3.6/tkinter/__init__.py", line 2017, in __init__
    self.tk = _tkinter.create(screenName, baseName, className, interactive, wantobjects, useTk, sync, use)
_tkinter.TclError: no display name and no $DISPLAY environment variable
```

small fix that worked on my cluster

### DIFF
--- a/umi_tools/umi_tools.py
+++ b/umi_tools/umi_tools.py
@@ -24,6 +24,8 @@ To use a specific tool, type::
 import os
 import sys
 import imp
+import matplotlib
+matplotlib.use('Agg')
 
 
 def main(argv=None):


IanSudbery:
Hi @Hoohm, Do you mind if I move this to whitelist.py? import of matplotlib can be slow, and its only used in one place.

IanSudbery:
ACtaully, might be easier if you did it, the git is a bit confusing on me doing it. 

The two lines are good - i've used theym before in other projects, if you could just put them at the top of `umi_tools/whitelist.py` rather than `umi_tools.py`, i'd be very greatful.

IanSudbery:
I'm wrong. Its used in `umi_methods`

PR #159, opened 2017-08-16
{is} remove tryexcept

IanSudbery:
Removes try except statements from around the import statements. 

I think these were neccessary because previously if you imported ran in a local directory that contained an `umi_tools` package, it would try to load from there and then get confused, which messed with the tests. This is fixed by adding `from __future__ import absolute_import`
also moved `matplotlib` backend specification to umi_methods.py from `umi_tools.py`

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -2,6 +2,6 @@
 include ez_setup.py
 # extensions
 include umi_tools/*.pyx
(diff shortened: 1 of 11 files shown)


TomSmithCGAT:
Feel free to merge if there are no more commits

PR #162, opened 2017-08-17
Changed default compression level from 9 to 6.

IanSudbery:
More than two fold speed improvement.

### DIFF
--- a/umi_tools/Utilities.py
+++ b/umi_tools/Utilities.py
@@ -506,12 +506,13 @@ def openFile(filename, mode="r", create_dir=False):
             if mode == "r":
                 return gzip.open(filename, 'rt', encoding="ascii")
             elif mode == "w":
-                return gzip.open(filename, 'wt', encoding="ascii")
+                return gzip.open(filename, 'wt', compresslevel=6,
+                                 encoding="ascii")
             else:
                 raise NotImplementedError(
                     "mode '{}' not implemented".format(mode))
         else:
-            return gzip.open(filename, mode)
+            return gzip.open(filename, mode, compresslevel=6)
     else:
         return open(filename, mode)
 
--- a/umi_tools/extract.py
+++ b/umi_tools/extract.py
@@ -73,6 +73,7 @@ Using regex and filtering against a whitelist of cell barcodes:
         -L extract.log [OPTIONS]
 
 '''
+from __future__ import absolute_import
 import sys
 import regex
 import collections


TomSmithCGAT:
I thought this was massively inflating the size of the gzipped files?

IanSudbery:
It basically produces files the  same size as commandline gzip

TomSmithCGAT:
Fair enough.

I think we should at least mention in the documentation that the gzip compression level is lower than python default and say that one can e.g pipe to `| gzip -9 - > out.gz` if higher compression is preferred.

My preference would be to include a `--comp-level` option in `extract` and default to 6 but this may be adding yet another option which will never be used?..

Happy with whatever you decide.

k3yavi:
Minor thing but if we can just dump plain text and pipe it ti https://zlib.net/pigz/ then we can utilize multithreads!

IanSudbery:
We are generally outputting more than one file. Would that work with this?

I think I compression option is reasonable

k3yavi:
Yea I don't think multiple files can be handled in this way.

TomSmithCGAT:
@k3yavi - The user can output uncompressed if they want and pipe it to their compression program of choice. E.g, for 10X data where one only wants to keep the second read pair, can use the following options with `extract`:

```
--log2stderr
--read2-out-only 
--read2-out=/dev/stdout 
```

As @IanSudbery says though, this doesn't work where you want to output both read pairs. Speaking of which, @IanSudbery do you think we should replace the `--read2-out-only` option (which stops output of read 1) with `--read2-stdout` (which would stop output of read 1 and output read 2 to stdout)? Seems cleaner

IanSudbery:
> @IanSudbery do you think we should replace the --read2-out-only option (which stops output of read 1) with --read2-stdout (which would stop output of read 1 and output read 2 to stdout)? Seems cleaner

Yes. 

I'll do that.

IanSudbery:
New commits include `--read2-stdout` option for `extract` and `--compresslevel` global option. 

@TomSmithCGAT Can you check this over.

TomSmithCGAT:
@IanSudbery, all looks good to me.

The scRNA-Seq guide will need a very minor update to reflect the modified options, but this branch came off master before the guide was added so I'll do this on the `{TS}-Add_OSX_Travis` branch.

Merge away...

IanSudbery:
I already updated the docs.

PR #170, opened 2017-08-21
debugs error when using --set-cell-number and --plot-prefix options t…

TomSmithCGAT:
…ogether


TomSmithCGAT:
Usiing `whitelist` options `--set-cell-number` and `--plot-prefix` together lead to unwanted error. This rectifies.

PR #177, opened 2017-09-05
Added new tool "tabulate" which output a table of all (grouped) UMIs and their read counts

fgp:
Hi,

This is kind of a hybrid of "group" and "count". For my project, I need to get the (summed) read-counts for each group of merged UMIs. "count" throws away too much information to be useful to me -- it reports only the per-gene total UMI count. "group", on the other hand, does too little -- it only groups reads, but would leave it to me to associate them to gene and to sum read counts over groups of reads with similar UMIs. Until now, I used a custom python script that did a job similar to what UMI-Tools does. To be able to switch over to UMI-Tools, I implemented the "tabulate" mode which outputs exactly what I need. It produces a table with one row per UMI-group, showing the gene and cell (if applicable), as well as the most common start and end mapping positions, total read count, and number of distinct UMIs in the group.

I'd love to see this integraded in UMI-Tools. This pull request contains a first stab at an implementation. It works, but isn't perfect yet. For one, it lacks documentation. And it needs to access both reads of a mate pair to correctly compute start and end mapping positions (because, like the rest of UMI-Tools, it attempts to take clipping into account). The lack of documentation I'll address soonish. For the 2nd problem I don't have a good solution yet. I'm sending this pull request not because I expect this to be merged in its current form (at least the documentation should be there, I gess), but because I wanted to have feedback on whether you'd consider merging this feature at all, and whether the approach I took is acceptable.

best,
Florian

### DIFF
--- a/umi_tools/tabulate.py
+++ b/umi_tools/tabulate.py
@@ -17,6 +17,7 @@ import sys
 import collections
 import re
 import os
+import itertools
 
 # required to make iteritems python2 and python3 compatible
 from builtins import dict
@@ -32,6 +33,12 @@ import umi_tools.umi_methods as umi_methods
 
 # add the generic docstring text
 __doc__ = __doc__ + U.GENERIC_DOCSTRING_GDC
+
+def most_common(L):
+    groups = itertools.groupby(sorted(L))
+    def _auxfun((item, iterable)):
+        return len(list(iterable)), -L.index(item)
+    return max( groups , key=lambda(x, v): (len(list(v)), -L.index(x)) )[0]
 
 def main(argv=None):
     """script main.
@@ -96,11 +103,17 @@ def main(argv=None):
 
     bundle_iterator = umi_methods.get_bundles(
         options,
-        only_count_reads=True,
+        all_reads=True, return_read2=True,
         metacontig_contig=metacontig2contig)
+
+    read2_stash = dict()
 
     for bundle, key, status in bundle_iterator(inreads):
         if status == "single_read":
+            read = bundle
+            if read.is_read2:
+                U.info("R2: %s" % read.qname)
+                read2_stash[read.qname] = read
             continue
 
         gene, cell = key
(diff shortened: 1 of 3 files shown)


TomSmithCGAT:
Hi Florian,

I suspect you should be able to get the output you require from group with a very simple parsing of the output using the `--group-out` option. I suggest we discuss this a bit further to see if this is the case before putting in the work to add another fully documented and tested command.

From what I gather from your comment above and the code as it stands for `tabulate`, you want to identify the groups of UMIs with the same gene and output the following for each UMI group:

- gene
- pos
- end
- umi ( I assume this is the the 'top' UMI)
- rawumis ( I assume this is the total number of unique UMIs, inc. errors)
- reads
- (+/-) cell where applicable

This information is contained in the output of `group --group-out`, with the exception of 'end' and 'cell', both of which could be easily added. All you would need to do then would be to parse the output of group and tally the number of unique UMIs and reads per group which could be done with a bash one-liner. 

Have I properly understood what you're trying to achieve?

fgp:
Hi Tom,

Thanks for your quick response.

Except for the missing "end" and "cell" columns, I guess `group --group-out` contains a superset of the information I need. It does, however, potentially contains much more -- in some of my data, UMI groups are comprised of many hundreds of reads, meaning that `group --group-out` will generate lots of actually unnecessary data. And some information, like the number of unique UMIs within a group, would have to be re-computed. This is particularly annoying because the next pipeline step is implemented in R, which is notoriously bad at stream processing of data, so I'd have to load the whole --group-out output at once (which requires lots of memory).

Maybe a compromise could be a third output file for the `group` command, which contains a single line per group? That could probably be done with a few tens of lines in group.py, and wouldn't required a whole new tool as my approach did.

Anyway, for now I'll try using `group --group-out`, and see how that goes.

Thanks,
Florian

TomSmithCGAT:
Hi Florian. It should be possible to pass the option `--group-out=/dev/stdout` and then parse the group output on the stdout with bash (or pass it to a python script). This would avoid saving out the full output of group. Do you need the "end" and "cell" columns?

fgp:
That /dev/stdout trick looks neat, I'll try that! Eventually to fully replace my existing pipeline I will need "end" and "cell" columns. But for now I can do without them. The cell column looks easy to add. For the end column, I think a clean solution would require extending bundle_iterator to collect and hand back both reads of a mate pair, instead of just the first.

TomSmithCGAT:
Yeah this is why I asked if you need it as it's not as straightforward as the 'cell' column. We can have a look into this soon. I'll raise this as an issue to keep track of progress

fgp:
It turned out that I reached a point where I need the "end" (actually, the "position" for read2) sooner that I had expected. I've now patched umi_tools to report that. The patch follows the basic idea outlined above, i.e. it extends bundle_iterator. I'll create a new pull request for that patch, and close this one. Hope that's how things are usually done, I haven't used github to contribute before...

TomSmithCGAT:
Hi @fgp - We're very happy to receive suggestions for improvement/modifications to the code base and this is definitely the right way to do it.

Issue #190, opened 2017-09-28
Division by zero in log message


koelling:
This is a bit of a weird edge case, but I was trying to run umi_tools dedup on a BAM file that only contained a single read pair, of which the second read is unmapped, and got this error:

```
Traceback (most recent call last):
  File "bin/umi_tools", line 11, in <module>
    load_entry_point('umi-tools==0.5.0', 'console_scripts', 'umi_tools')()
  File "lib/python3.5/site-packages/umi_tools/umi_tools.py", line 50, in main
    module.main(sys.argv)
  File "lib/python3.5/site-packages/umi_tools/dedup.py", line 391, in main
    processor.UMIClusterer.positions))
ZeroDivisionError: float division by zero
```

I think this is caused by this line, where you divide by processor.UMIClusterer.positions, which is zero in this case: https://github.com/CGATOxford/UMI-tools/blob/master/umi_tools/dedup.py#L390

IanSudbery:
Good catch. Should be an easy fix. I guess.

slebedeva:
Got the same error when working with test data (just a small part of the original fastq).

TomSmithCGAT:
Hi @koelling & @slebedeva - Would you mind installing from the {TS}-Debug_no_positions branch to confirm this bug is resolved on the branch?

slebedeva:
Confirmed - the bug is gone for me! Thanks a lot!!!

TomSmithCGAT:
Brilliant. I'll merge this branch with master and release 0.5.1 shortly

TomSmithCGAT:
Resolved in version [0.5.1](https://github.com/CGATOxford/UMI-tools/releases/tag/0.5.1)

PR #195, opened 2017-10-13
{ts} debug no positions

TomSmithCGAT:
Resolves (#190) as far as I can tell. Have confirmed that `dedup` previously errored with non-informative message when input BAM containing no reads. This is no longer the case. If there are no positions for de-duplication, this is now raised as a warning in the logfile but exit status is 0.


IanSudbery:
Is the stuff in `umi_methods` adressing the same or a different problem?

Anyway, all looks good to me.

TomSmithCGAT:
Oh sorry, forgot to mention that. Different problem relating to a (very unusual) case where I wanted to ignore the UMI but use the cell barcode. Thought I'd kill two birds with one stone and get both these debugs merged in together

Issue #185, opened 2017-09-13
count confusingly uses only the umi as grouping key and ignored the mapping position


fgp:
From the documentation, I expected `count` to use the same grouping algorithm as `dedup` and `group`, but to then simply count the number of distinct (post-grouping) UMIs per gene (identified either by tag or RNAME).

But it seems that`count` will merge similar UMIs within a gene regardless of the mapping position, contrary to how `dedup` and `group` operate. 

I this by design, or is this a a bug? I have to admit that the intended interplay between `--per-gene`,`--per-contig` and `--whole-contig` (which exists only for group and dedup it seems) isn't really clear to me.

TomSmithCGAT:
Hiya @fgp -  This is intended. `count` is currently designed to only handle 3'-Seq data in which reads from the same unique molecule can have different alignment positions since fragmentation occurs after the amplification. This is explained in the [single_cell_tutorial](https://github.com/CGATOxford/UMI-tools/blob/master/doc/Single_cell_tutorial.md#step-5-assigning-reads-to-genes) but I agree the documentation for count is woefully inadequate about this!

If you run `count`, `dedup` or `group` with `--per-gene` you will get this behaviour, e.g ignore position, use just gene assigned. In essence the output of each tool going from group -> dedup -> count is just one step further along the process of counting unique reads per gene. The `--per-contig` and `--gene-tag` options are used to define whether to tag the assigned gene from a gene tag (e.g after `featureCounts -R`) or from the alignment contig (e.g if reads are aligned to a transcriptome).

`--whole-contig` is a completely separate option which forces `umi_tools` to read in every read per contig before defining UMI groups. By default, it processes along the contig and when it has reached 1000bp beyond the last read in a group, it yields the group for deduplication. This is OK for short reads but there may be instances where the user wants to be sure no reads are missed from a group, which is what `--whole-contig` is for. This comes at a significant memory cost however.

IanSudbery:
Just to add...

[...]

PR #196, opened 2017-10-13
renames --whole-contig option to --buffer-whole-contig and updates doc.

TomSmithCGAT:
The naming of this option causes confusions since the introduction (by me) of very similar sounding `--per-contig` option. E.g see #185. Hopefully, the renaming will resolve this confusion


IanSudbery:
Looks good.

PR #267, opened 2018-08-01
Suggest installation via bioconda instead of custom channel.

johanneskoester:
This also increases interoperability with other tools.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -69,7 +69,7 @@ If you're using Conda, you can use:
 
 .. code:: bash
 
-   $ conda install -c https://conda.anaconda.org/toms umi_tools
+   $ conda install -c bioconda -c conda-forge umi_tools
 
 Or pip:
 


TomSmithCGAT:
I'm happy with this. @IanSudbery?

IanSudbery:
Is the version on bioconda up-to-date? If so then we should definately be recommending bioconda.

johanneskoester:
Yes, it is the same as in your channel, and you are very welcome to maintain the recipe in bioconda (either via pull requests from a fork, or via joining your team).

johanneskoester:
xref: https://anaconda.org/search?q=umi_tools

johanneskoester:
We also have automatic testing and a convenient build infrastructure for our recipes, so maintenance will even be easier for you.

IanSudbery:
Thanks @johanneskoester, I definately think that it is better to not do things on custom channels.

PR #271, opened 2018-08-13
added `__version__`

jbloom:
Having a `umi_tools.__version__` string
is the standard recommended Python way
to make the version number accessible.
This commit does that.

### DIFF
--- a/umi_tools/__init__.py
+++ b/umi_tools/__init__.py
@@ -1,1 +1,3 @@
+from __future__ import absolute_import
+
 from umi_tools.version import __version__
--- a/umi_tools/__init__.py
+++ b/umi_tools/__init__.py
@@ -1,0 +1,1 @@
+from umi_tools.version import __version__


TomSmithCGAT:
Hi @jbloom. Forgive my ignorance on how best to implement python package versions. It seems like there are many ways to maintain a single source of truth for the version number and . Currenty, we're storing the version number in the `__version__` variable `umi_tools/version.py` and importing this where required by modifying `sys.path`, e.g when the user call [`umi_tools --version`](https://github.com/CGATOxford/UMI-tools/blob/master/umi_tools/umi_tools.py#L44-L51). I have no idea if this is the most pythonic way to do this but it does look like this would be more [PEP8 compliant.](https://www.python.org/dev/peps/pep-0396/)

As I understand it, your PR adds an import to `__init__.py` to enable import in other scripts via  `from umi_tools import __version__`, but the places where `__version__` is used as still using the current method of importing via modification of `sys.path`. Is there any reason for this? Also, this is currently [failing the python 2 tests ](https://travis-ci.org/CGATOxford/UMI-tools/jobs/415671846) which I guess is due to the differences in how p2 and p3 set `sys.path`?
```
from umi_tools.version import __version__
ImportError: No module named version```

TomSmithCGAT:
Brilliant. Thanks!

Issue #276, opened 2018-09-20
Missing CB or UB with umi-tools count


dhb2128:
Hi, thanks for the useful software!

I'm trying to recount a bam file generated via 10x, but I'm getting errors since my bam file does not have the 10x-corrected cell-barcode or umi-barcode tag for every read (CB and UB, respectively)

```
FILE="neurons_900_possorted_genome_bam"
umi_tools count \
--extract-umi-method="tag" \
--per-gene \
--per-cell \
--umi-tag=UB \
--cell-tag=CB \
--mapping-quality=255 \
--gene-tag=GN \
--skip-tags-regex=";" \
--method="unique" \
--per-cell -I $FILE.bam -S counts.tsv.gz
# output generated by count --extract-umi-method=tag --per-gene --per-cell --umi-tag=UB --cell-tag=CB --mapping-quality=255 --gene-tag=GN --skip-tags-regex=; --method=unique --per-cell -I neurons_900_possorted_genome_bam.bam -S counts.tsv.gz
# job started at Wed Sep 19 20:44:31 2018 on compute-a-16-165.o2.rc.hms.harvard.edu -- eb361b58-51f6-4ff6-aa62-699d3d743044
# pid: 171613, system: Linux 3.10.0-327.3.1.el7.x86_64 #1 SMP Wed Dec 9 14:09:15 UTC 2015 x86_64
# cell_tag                                : CB
# cell_tag_delim                          : None
# cell_tag_split                          : -
# chrom                                   : None
# compresslevel                           : 6
# gene_tag                                : GN
# gene_transcript_map                     : None
# get_umi_method                          : tag
# ignore_umi                              : False
# in_sam                                  : False
# log2stderr                              : False
# loglevel                                : 1
# mapping_quality                         : 255
# method                                  : unique
# paired                                  : False
# per_cell                                : True
# per_contig                              : False
# per_gene                                : True
# random_seed                             : None
# short_help                              : None
# skip_regex                              : ;
# stderr                                  : <_io.TextIOWrapper name='<stderr>' mode='w' encoding='UTF-8'>
# stdin                                   : <_io.TextIOWrapper name='neurons_900_possorted_genome_bam.bam' mode='r' encoding='UTF-8'>
# stdlog                                  : <_io.TextIOWrapper name='<stdout>' mode='w' encoding='UTF-8'>
# stdout                                  : <_io.TextIOWrapper name='counts.tsv.gz' encoding='ascii'>
# subset                                  : None
[...]

PR #281, opened 2018-09-26
{ts} debug missing tag

TomSmithCGAT:
This enables reads missing a UMI or CB tag to be skipped rather than throwing an error (See #276).

As part of this, I've also moved the check for UMI/CB tags to before the check for bundle yielding as per all the other checks which cause reads to be skipped.


TomSmithCGAT:
@IanSudbery - nudge nudge

ZHUwj0:
Hi @TomSmithCGAT,

Today, I just downloaded UMI-Tools: Version 1.0.0, and try to use it to do umi_tools dedup on BAM files from cellranger (10X genomics data).   However, I still have the error, "KeyError: "tag 'CB' not present". 
I noticed that you actually fixed this issue before.  is it possible for you to check whether this bug had been fixed or not, again?   
many thanks. 

Best regards, 
Wenjuan

ZHUwj0:
For example,  NH:i:1  HI:i:1  AS:i:69 nM:i:9  RE:A:I  xf:i:0  li:i:0  CR:Z:CCGTAGACACTGTGTA   CY:Z:FFFFFFFFFFFFFFFF   UR:Z:TCAGGAACCG UY:Z:FFFFFFFFFF UB:Z:TCAGGAACCG RG:Z:1823_BA24:0:1:HCC5FDMXX:1

TomSmithCGAT:
Hi Wenjun,

by default, UMI-tools expects the cell barcodes to be provided in the 'CB' tag. If you have another tag denoting the cell barcode, you can add the option `--cell-tag=CELL_TAG`, where `CELL_TAG` appears to be `CR` in your example above

ZHUwj0:
Hi @TomSmithCGAT,
yes, I knew this. Here is my command:
umi_tools dedup --per-cell --output-stats=sample --stdin=test.bam --stdout=test.dedup.bam --extract-umi-method=tag --umi-tag=UB --cell-tag=CB. 

However,  In BAM files from cellranger standard pipeline,  the majority of lines have the CB tag.  But, some lines do not have CB tag, like this one. 
NH:i:1 HI:i:1 AS:i:69 nM:i:9 RE:A:I xf:i:0 li:i:0 CR:Z:CCGTAGACACTGTGTA CY:Z:FFFFFFFFFFFFFFFF UR:Z:TCAGGAACCG UY:Z:FFFFFFFFFF UB:Z:TCAGGAACCG RG:Z:1823_BA24:0:1:HCC5FDMXX:1

Possibly that is why umi_tools had this error, "KeyError: "tag 'CB' not present". 

Best regards, 
Wenjuan

TomSmithCGAT:
Ah I see. Sorry, I misunderstood the problem. I'll take another look at this now

TomSmithCGAT:
I can't see any obvious reason why this would happen. This should be caught here: https://github.com/CGATOxford/UMI-tools/blob/6945f6ae5a9a7c56fdac22f2caddec4246442bc2/umi_tools/sam_methods.py#L411-L427

 Would you mind posting the full error message here so I can see where the KeyError originates and confirming your version number (`umi_tools --version`)

ZHUwj0:
Please check this file including the error and real example of BAM file. 
[test.zip](https://github.com/CGATOxford/UMI-tools/files/3931643/test.zip)

You may want to use this small BAM file to test software. 

many thanks.

TomSmithCGAT:
Ah, right, got it! The problem is that the KeyError originates from the parsing of the BAM to generate the null distribution for the stats (code below) rather than the parsing for deduplication (code above). I'll add a patch for this now.

https://github.com/CGATOxford/UMI-tools/blob/5e99b633932d7de7591cd9881db8745295ccb500/umi_tools/umi_methods.py#L164-L172

TomSmithCGAT:
Hi Wenjuan,

Could you test the version of UMI-tools on the `{TS}-DebugCellTag` branch. This works with the test BAM so should be fine but just want to be sure before pushing changes to the master.

Thanks,

Tom

ZHUwj0:
Hi Tom, 

After testing it,  this version of UMI-tools works very well.  Thank you so much for your help. 

Wenjuan

TomSmithCGAT:
Great. The branch has been merged into master. Thanks for bringing this to our attention and for providing me with the example BAM.

PR #283, opened 2018-09-28
Use correct option's name for --bc-pattern2

reineckef:
Also added the command-line name to the error messages.


TomSmithCGAT:
Thanks @reineckef!

PR #299, opened 2018-12-08
optimized getErrorCorrectMapping() with bktrees

redst4r:
Hi,

`umi_methods.getErrorCorrectMapping()` can be sped up drastically using a BK-Tree as data-structure for storing the whitelist of cell-barcodes. 

- the old version used pairwise comparisons to find whitelisted barcodes  1BP apart from a query.
- now we 'index' the whitelist using a BKTree, which makes retriving 1BP sequences alot faster
- depends now on the pybktree package available in pip


TomSmithCGAT:
Thanks for this very helpful update @redst4r and apologies for the delay looking over it. It all looks OK to me and the error correction is covered by the regression tests so I'm happy that the output is consistent.

@IanSudbery - Are you happy for this to be merged?

umi_tools/umi_methods.py
```diff
@@ -358,7 +358,7 @@ def getKneeEstimate(cell_barcode_counts,
     return final_barcodes
 
 
-def getErrorCorrectMapping(cell_barcodes, whitelist, threshold=1):
```
TomSmithCGAT: Please delete the previous `getErrorCorrectMapping` function

IanSudbery:
Seems good in principle, but:

* It means relying on a package not available in `conda`
* Although the error correction is covered in the tests, the tests are failling. 

I don't know if this is a problem with the content of the tests or with the Travis environment

@redst4r: have you run the tests on your system? Do they pass?

TomSmithCGAT:
@IanSudbery - Most recent travis build passed unless I'm mistaken (https://travis-ci.org/CGATOxford/UMI-tools/builds/472480700?utm_source=github_status&utm_medium=notification). Do you mean the tests failed for you locally?

I hadn't picked up on the `conda` issue. I guess we could either 1) get `pybktree` into bioconda or 2) use `pip` in the conda recipe?

TomSmithCGAT:
@redst4r, have you worked with `conda` before and if so, would you be able to add `bktrees` to the `bioconda` channel. This would help to keep `umi_tools` easy to install.

redst4r:
Hi guys,

@IanSudbery yeah I noticed some unrelated tests failing. On my machine the tests already failed on `master` so no idea what was going on there. Seems to have resolved magically now...

@TomSmithCGAT yeah I'll get rid of the old code!

I was thinking about adding `pybktree` to bioconda. Never done it before, but it doesnt seem hard. Only issue: They state that only "bioinformatics"-related packages should be added to conda-recipies. 

The only real issue (coding-wise): `pybktree` somehow doesn't work with the cython-based distance function you guys implemented (tests fail if I use the cython-distance function), so i replaced it with my own python-hamming-distance. 
Could be a problem of the string->bytes conversion

TomSmithCGAT:
Hi @redst4r. How are you getting on with adding `pybktree` to bioconda? I'm happy to do this if you haven't had a chance to do it yourself.

Hoohm:
@redst4r 
quick question regarding this line in your code:
`candidates = [white_cell for d, white_cell in barcode_tree.find(cell_barcode, 1) if d>0]`

What is the d>0 for?

I tested a few lists but I couldn't find a difference with or without it.

redst4r:
sry didnt have time to look into the whole thing ever since!

@Hoohm yeah, seems like the ` if cell_barcode in whitelist: continue` should already take care of that. `d==0` would only happen if the `cell_barcode` matches a member of the bktree exactly, hence it was part of `whitelist` in the first place

TomSmithCGAT:
I've added `pybktree` to `conda-forge`: https://github.com/conda-forge/pybktree-feedstock

TomSmithCGAT:
Hi @redst4r - Apologies for the long delay getting this PR merged. In the meantime, we've moved the `getErrorCorrectMapping` function to `umi_tools/whitelist_methods.py` which is nowcausing a conflict with the merge with master. Would you mind updating your branch and I'll get this merged pronto. Thanks!

TomSmithCGAT:
@IanSudbery - Any objections to this being merged into master?

IanSudbery:
Erm. EIther I'm looking at this wrong, or in its current state, this just import pybktree, but doesn't do anything with it?

TomSmithCGAT:
Yeah, I'm an idiot. In the conflict resolution, I managed to remove the updated `getErrorCorrectMapping` using `pybktree`! 
I've added this back in. Now tests that were introduced since PR was submitted are failing. I'll look into this...

TomSmithCGAT:
In digging into the test fails, I discovered a bug in the current code. Currently `getErrorCorrectMapping` uses the `edit_distance` function without any check that CB A and B are the same length, which is also not checked within `edit_distance`. The way the later is written, it returns different distances depending on the order of comparison.

Because of the above bug, 3 of our whitelist test outputs contained the line
TGACCA**T**TACTTGT**AA**AGG	TGACCA**G**TACTTGT**GT**	3	1
where column 2 contains a 16nt CB within the error threshold (3) of the whitelisted 19nt CB in column 1. The 3 substitutions are in bold and because of the direction of the comparison, the last 3 nt in column 1 CB are ignored. This was the only line causing the branch to error.

On the UMI side, we demand the UMIs to be the same length, so the above oversight in `edit_distance` is not a concern. Nevertheless, I've added a check within the function so that it returns `np.inf` if the barcodes are not the same length.

IanSudbery:
In terms of different length CBs, I know that some people are using protocols that generate different length CBs, although I think they usually think this is due to an indel. Here were are saying that two CBs of different lengths are definitely different. Is that the correct choice? I'm not sure really what the alternative would be.

IanSudbery (commented):
I'm good with this as long as we are good with the decision that two CBs of different lengths are defaintely different. I'm not sure there is a time-reasonable alternative, but we should be aware that is the decision we are making.

Issue #138, opened 2017-06-19
Filter whitelisted barcodes within edit distance of another whitelisted barcode


TomSmithCGAT:
Based on my analysis of droplet scRNA-Seq cell barcodes (e.g blog post [here](https://cgatoxford.wordpress.com/2017/05/23/estimating-the-number-of-true-cell-barcodes-in-single-cell-rna-seq-part-2/)), I think we should add the option to remove cell barcodes within the automatically generated whitelist if they are within an edit distance threshold of another whitelisted barcode with greater frequency. I believe there is sufficient evidence to suggest error barcodes (INDEL or sequencing error) may pass the whitelist threshold. We could merge these barcodes into the true barcode from which they derive but this risks merging two truly different cells. On balance, removing these potential error barcodes seems like the best approach. This is compatible with the current error correction within extract which is restricted to only barcodes not in the whitelist. Thus the steps for whitelist generation and filtering would be:

1. Parse first 50M reads, extract cell barcodes and generate a whitelist using the knee method
2. (Optionally) identify all cell barcodes within an edit distance threshold of exactly one whitelisted barcode
3. (Optionally) Remove whitelisted barcodes within an edit distance threshold of another whitelisted barcode within greater frequency
4. Parse all reads, extract cell barcodes and filter reads against the whitelist (with optional correction of cell barcodes not in the whitelist)

TomSmithCGAT:
This is now available on the master branch and will be in the next release

PR #309, opened 2019-01-24
{ts} whitelist filter pos errors

TomSmithCGAT:
- Adds options to whitelist to detect putative errors in the CBs above the knee threshols
- Adds tests to cover new options

See (#138) for motivation


umi_tools/whitelist.py
```diff
@@ -42,14 +42,39 @@
...
+the log file (--log) will detail the number of reads from CBs above
+the threshold which may be errors. In most cases, we expect the number
+of reads to be a very small fraction of the total reads and therefore
+recommend taking the conservative approach. See
+https://cgatoxford.wordpress.com/2017/05/23/estimating-the-number-of-true-cell-barcodes-in-single-cell-rna-seq-part-2/
+for an analysis of errors in barcodes above the knee threshold.
+
+Optional detection of putative error CBs above the threshold can be
+switched on with the error detection option:
+--ed-above-threshold
```
IanSudbery: Why two different options? Why not `--ed-above-thershold=[discard|correct]` with a default setting of False/None?

umi_tools/umi_methods.py
```diff
@@ -438,6 +440,70 @@ def getUserDefinedBarcodes(whitelist_tsv, getErrorCorrection=False):
...
+    '''
+    Check for errors (substitutions, insertions, deletions) between a barcode
+    and a set of whitelist barcodes.
+
+    Returns the whitelist barcodes which match the input barcode
+    allowing for errors. Returns as soon as two are identified.
+    '''
+
+    near_matches = []
+    comp_regex = regex.compile("(%s){e<=%i}" % (barcode, errors))
```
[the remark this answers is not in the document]
TomSmithCGAT: Yes, there must be. I'll look into this.

umi_tools/whitelist.py
TomSmithCGAT: Yeah, that would be better. One other issue is what error threshold to allow for the error detection above the threshold. Right now, it re-uses the `--error-correct-threshold` which is still described as "Hamming distance for correction of barcodes to whitelist barcodes", e.g same error threshold correcting 1) below knee to above knee and 2) between the above knee barcodes. I assume this is the behaviour that users will desire but maybe there will be an edge case where different thresholds are desirable. For now, I'll just update the description for `--error-correct-threshold` to make this clearer.

TomSmithCGAT:
As suggested, I've removed the `ed-resolution` option and just left the option `ed-above-threshold=[discard/correct]`, defaulted to `None`.

I've also implemented solution 1 above. E.g putative INDEL CBs are always discarded. The discarding/correcting behaviour is logged like so. Note the below is from the testing where we allow 3 errors in a 16-20 base CB in order to actually detect some "error" CBs. Hence the relatively large number of putative CB errors detected.

```
2019-02-06 15:52:02,654 INFO Top 414 cell barcodes passed the selected threshold
2019-02-06 15:52:02,991 INFO CBs above the knee corrected due to possible substitutions: 6
2019-02-06 15:52:02,991 INFO CBs above the knee discarded due to possible INDELs: 26
2019-02-06 15:52:02,991 INFO CBs above the knee discarded due to possible errors from multiple other CBs: 0


```

TomSmithCGAT:
As far as I'm aware, this branch is now good to merge?

IanSudbery:
okay, good to go

Issue #175, opened 2017-08-24
How to handle paired end reads with barcodes on both reverse and forward reads?


J-sara:
Hello,

I have a paired end fastq file and my experiment is designed in a way that each PAIRED READ has ONE barcode in it. This barcode might be on the forward  read or on the reverse read (not both of them) and the barcode has a specific sequence before and after it that helps to identify the barcode. So, some of the forward reads have the barcode and some of them do not. This is also true for reverse reads. 

My question is whether UMI tools is able to handle this type of experiment design or not and if so, how I can do it?

Best

TomSmithCGAT:
Hi @J-sara - Unfortunately, UMI-tools will not be able to handle this as it expects the read ids to be identical, with both containing the UMI. In fact, as I understand it,  it's conventional for read pairs to have the same read id although I can't see any reference to this in the Sam file format documents. 

I'm assuming from the experimental design you describe that the barcodes are on one read pair or other at random in the fastq file? If this is the case, you'll need to write a bespoke script to extract the barcode from whichever read it resides in and append it to both reads ids in the format expected by UMI-tools `dedup`, e.g separated from the read_id by a "_".

As far as I'm aware, there are no tools which could handle the UMI being on one read pair or the other at random in the BAM file.

IanSudbery:
@TomSmithCGAT I didn't read @J-sara 's post as saying they were on one or the other read name, but on one or other read sequence. 

Could the regex pattern matching not help us here? What happens with the regex if it fails to match on a read?

TomSmithCGAT:
@IanSudbery - My bad. I just assumed we were talking about using e.g UMI-tools `dedup` on a BAM file. 

If the regex fails to match with `extract`, the read pair will be discarded. This is intentional behaviour. E.g for inDrop where there is an adapter sequence one wants to match.

TomSmithCGAT:
Hi @J-sara - Did you manage to write a bespoke script to extract the UMIs. If not, I can help you out if you can share some more information about how the barcodes are encoded and a test fastq

OldManTae:
Hi @TomSmithCGAT - I have a similar issue to @J-sara, in that the barcode I am looking for can appear in either Read 1 OR Read 2. Which one it appears in can be determined by the presence of a specific sequence immediately after the barcode.

I had a couple of questions regarding the options for extract (with my guesses in brackets):

1. does bc-pattern in paired end reads search specifically in read 1? (Yes)
[...]

PR #311, opened 2019-01-25
{ts} barcode extract either read

TomSmithCGAT:
Note, the commits on Jan 24 are from a merge with branch `{TS}-EnablePerCellCountTab` (PR #308). Probably best to merge that PR before reviewing the changes herein.

See #175 for motivation
- Implements barcode extraction from read1 *or* read2 with `--either-read`. 
- Implements option to either discard reads where both match, or keep the UMI with highest sequence quality
- Adds test for `--either-read`

I don't have a good input sample for the test so the current input is hacked together from the indrop data (see `tests/README`). Basically swapped read1 and read2 at random and use option to identify which read matches the regex. Of course, the output from this is now garbage. I'll post on #175 and ask for more suitable toy data. Also need to add a test to cover `--either-read-resolve=quality`


umi_tools/extract.py
```diff
@@ -190,77 +219,19 @@ def main(argv=None):
...
-                        options.pattern, options.pattern2))
-        elif options.extract_method == "regex":
-            U.error("barcode regex(es) do not include any umi groups "
-                    "(starting with 'umi_') %s, %s" (
-                        options.pattern, options.pattern2))
+    extract_cell, extract_umi = U.validateExtractOptions(options)
+
+    if options.either_read:
+        if extract_cell:
+            U.error("Option to extract from either read (--either-read) "
```
IanSudbery: Why not?

TomSmithCGAT: Quite simply because I'm not sure what we would want to do with the cell barcode. Right now, we're supporting the most basic form of barcodes on either read: 

- There is a single UMI
- This could be on either read
- The correct read can be identified using regexes
- If both reads match their regex, either discard read pair, or designate one read the UMI read based on base call qualities

We could extend this to allow cell barcodes to be extracted in the same manner but I'm not sure what we're expecting? Do we expect the cell barcode and UMI are on opposite reads but could be either way around, or maybe they are on the same read? Perhaps the cell barcode is on both reads but the UMI just one, or vis versa? 

My preference would be to not support cell barcodes within the "either read" option until we know we're dealing with. Happy to put a suggestion in the error message to raise an issue on GitHub should any issue desire this functionality.

IanSudbery: Okay, lets see if anyone complains.

TomSmithCGAT:
@IanSudbery  - OK for merge?

IanSudbery:
How about allowing the user to just provide one regex if it can be on either read?

TomSmithCGAT:
Oops, sorry, missed this! 

We could have a single regex option by simply reusing `--bc-pattern` if `--bc-pattern2` is not provided (I assume this is what are thinking). Personally, I'd rather force the user to be explicit so there's no unexpected behaviour if they forget to specify `--bc-pattern2` if it is different.

IanSudbery:
go ahead and merge this.

Issue #304, opened 2019-01-16
High memory usage


Poshi:
I'm running umi_tools group (v0.5.3) on a bam file of 1.1TiB containing 3401822722 records. On a 256GiB machine, its memory usage is peaking at around 147GiB (and the rest is being used for file cache, mostly). Does it make sense to you? It looks like a huge amount of memory for such task.

If I'm not misunderstood, all reads aligned in every position have to be considered at once, but I would be surprised if the maximum number of reads in a single position is so high to get to those numbers.

Is there any way to know the memory requirements of an execution in advance? It would be good for the proper scheduling in our queue system.

TomSmithCGAT:
Hi @Poshi. This does sound like a huge memory usage but I've never applied `umi_tools group` to a 3.4B read input(!) so I've really no handle on what to expect. Unfortunately, there's no way to know in advance what the run time or memory usage will be as it depends on the distribution of unique UMIs per position. 

`umi_tools dedup` will log the maximum number of unique UMIs in a single position but `group` currently doesn't log this. I'll add this to the on-going minor update to documentation. In the meantime, would you be able to re-run using `dedup` so we can see what the max UMI count per position is.

IanSudbery:
The biggest determinants of the memory usage are:
    * Number of unique UMIs at a position. 
    * The interconnectivity of the UMI graph.

The read depth of the file, per-se does not make much of a difference, although obviously a deeper file will have a more interconnected UMI graph. 

For a fixed network saturation level (that is the % of all possible UMIs that are found at one location), and assuming that UMIs are picked more or less randomly, memory usage scales as O(n^2), where n is the maximum number of UMIs at a location in the file.

However, there is also the network saturation. We notice that when saturation gets to around 30% there is a large jump in time and memory usage. This is also the point at which it starts to become difficult to disentangle what is and error and what is a genuinely unique UMI - that is, it becomes likely that you will select two UMIs only one edit difference just by chance. i.e. at this point your UMI is probably too short to allow intelligent deduplication.

IanSudbery:
You'll find you use much less memory is you don't output-stats - turns out this is a big memory drain.

If you think you need the stats, we recommend only using a subset of your input to generate the stats and run the full file separately.  You can pass the option `--chrom` (try chr22 for example) to limit to one chromosome, or `--subset` to only use a fraction of reads.

Poshi:
[...]

PR #312, opened 2019-01-25
Deal with chimeric read pairs

TomSmithCGAT:
See  #304 for motivation

- Hardcodes `dedup` to discard chimeric read pairs
- Defaults `group` to discard chimeric reads. These can be output with `--output-chimeric` but will not be grouped. 

The above makes the most sense to me since the `TLEN` of chimeric read pairs is 0. Thus, the current approach in UMI-tools to identify groups of read pairs with the same alignment coordinates will not work for chimeric reads. Therefore these reads are discarded by default but can be output in `group` without any group information. This is the same behaviour as can be switched on for unmapped reads (`--output-unmapped`)

Of course, if we switched to using the actual alignment coordinate of read2 (soft-clipped corrected, or not) to determine the read groups in future, we could then retain chimeric read pairs.

### DIFF
--- a/umi_tools/Utilities.py
+++ b/umi_tools/Utilities.py
@@ -1070,6 +1070,32 @@ def validateSamOptions(options, group=False):
                              "to retain unmapped without deduplicating them, "
                              "use the group command")
 
+    if options.paired:
+        if options.chimeric_pairs == "use":
+            warn("Chimeric read pairs are being used. "
+                 "Some read pair UMIs may be grouped/deduplicated using "
+                 "just the mapping coordinates from read1."
+                 "This may also increase the run time and memory usage. "
+                 "Consider --chimeric-pairs==discard to discard these reads "
+                 "or --chimeric-pairs==output (group command only) to "
+                 "output them without grouping")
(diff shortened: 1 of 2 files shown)


TomSmithCGAT:
Good point. On reflection, deduplication of chimeric read pairs is no worse than single end deduplication in the sense that with SE sequencing, you're simple blind to the possible chimeric nature of the sequenced cDNA fragment.

Perhaps the best approach is get rid of the `--output-chimeric` option and have a `--chimeric-pairs` option with choices {`discard`, `use`, `output`}, where `use` = don't do anything special with chimeric read pairs, `TLEN` will be 0, and `output`=write them out but without any grouping. We could default to `use` to keep the output consistent but warn users about the issues with chimeric read pairs?

If you're happy with this, I'll also change the `--output-unmapped` option to `--unmapped-read={discard, output}` to try and make the options clearer. We can keep the `--output-unmapped` option for legacy purposes but suppress it from the help message and warn users it may be removed in future.

IanSudbery:
sounds good.

PR #313, opened 2019-01-31
{ts} add faq

TomSmithCGAT:
- Adds FAQ


IanSudbery:
I've added a list of known regex for different techniques we've come across. I felt like i've seen hundreds of these things, but when I actaully went to look through the issues and my emails, I couldn't find that many with technique names attached to them. I've also added a list of techniques that need `--per-gene`, almost certainly not exhaustive.

I kinda feel like:
1) We should have a seperate page explaining how to do the regexes. In particular, the named group syntax is apparent a) specific to python b) not very well known even within python.  I'll try to do this this weekend.

2) We should move most of the detailed `--help` text to on-line man pages. Again, I'll have a look at this at the weekend if I can. 

3) I should do a page explaining the API to the algorithm.

4) We shuold put all this together in a proper webpage. I know I said this last time. I'll take a look again. I actaully did it for iCLIPlib and readthedocs, and I seem to remember it wasn't too painful.

TomSmithCGAT:
Thanks for the update @IanSudbery. I agree with all the points above. I think 1,2 & 4 are probably the priority but 3 would be brilliant too. 

1. I hadn't realised named groups in regexes were non-standard. I'd also add c) We use `regex` not `re`. So lots of good reasons to make a specific page to explain how they are utilised in`umi_tools`.

TomSmithCGAT:
Regex guide reads very well 👍

TomSmithCGAT:
OK, so the main manual pages are now complete, with all options covered (I think). The remaining things to do for this PR are:

- [ ] Write a "common options" page and like to this in all tools pages

TomSmithCGAT:
@IanSudbery - We have 1,2 & 4 from above covered. Do you think we need 3 in this release?

PR #319, opened 2019-02-12
Refactored UMIClusterer to remove unncessesary arguement to __call__

IanSudbery:
Call signature for `UMIClusterer` was `__call__(umis, counts, threshold=1)`  where `umis` was a list of UMIs and `counts` was a dictionary mapping UMIs to their counts. This seemed unnecesssary as `umis` could be recovered from `counts` with `counts.keys()`. When I wrote the API website page, I pretended it wasn't there. This pull makes the reality match the docs.

Also fixes pysam<0.14 errors if tag is missing.

### DIFF
--- a/umi_tools/sam_methods.py
+++ b/umi_tools/sam_methods.py
@@ -411,9 +411,9 @@ class get_bundles:
                 try:
                     umi, cell = self.barcode_getter(read)
                 except KeyError:
-                    error_msg = "Read skipped, missing umi and/or cell tag"  
+                    error_msg = "Read skipped, missing umi and/or cell tag"
                     if self.read_events[error_msg] == 0:
-                        
+
                         # pysam renamed .tostring -> to_string in 0.14
                         # .tostring requies access to the parent AlignmentFile
                         try:
--- a/umi_tools/sam_methods.py
+++ b/umi_tools/sam_methods.py
@@ -411,10 +411,18 @@ class get_bundles:
                 try:
                     umi, cell = self.barcode_getter(read)
                 except KeyError:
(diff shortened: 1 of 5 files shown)


umi_tools/sam_methods.py
```diff
@@ -411,10 +411,18 @@ def __call__(self, inreads):
...
-                    error_msg = "Read skipped, missing umi and/or cell tag"
+                    error_msg = "Read skipped, missing umi and/or cell tag"  
                     if self.read_events[error_msg] == 0:
+                        
+                        # pysam renamed .tostring -> to_string in 0.14
+                        # .tostring requies access to the parent AlignmentFile
+                        try:
+                            formatted_read = read.to_string()
+                        except AttributeError:
+                            formatted_read = read.query_name
```
TomSmithCGAT: This is just the name of the read though, right? Ideally  `formatted_read` is the full BAM entry as a string. Does `str(read)` for all pysam versions by any chance, e.g does pysam set the `.__str__()` method?

TomSmithCGAT: If it's not possible to return the complete entry as a string in all versions of pysam, happy for pysam<0.14 to just return the read name

TomSmithCGAT:
Just the one minor comment

TomSmithCGAT:
Looks like `pep8` is unhappy though...

umi_tools/sam_methods.py
IanSudbery: In pysam < 0.14, the read object doesn't contain all the information neccessary to convert the read to text.

IanSudbery:
bah.

Issue #328, opened 2019-03-19
output rejected reads during whitelist


julietchen:
Hi,
I was wondering if there was a way during the whitelist call to output a fastq for the rejected reads (reads that don't match the cell barcode pattern). The reason for this is to look at why reads aren't matching the barcode pattern in order to improve things in the lab. Apologies if I've missed it somewhere in the documentation -- I've been looking around on github and readthedocs and can't quite find the answer I'm looking for.

I'm currently using version 0.5.5 but am willing to upgrade to v1.

Thanks!
Julie

TomSmithCGAT:
Hi Julie. No, this isn't currently possible but I can see the justification for doing so and should  be easy to add. Just to check, the idea would be to output *all* rejected reads?

julietchen:
Hi Tom,
Yes, exactly -- the idea is to take all the rejected reads and then group
and sort to see if there are any patterns (contaminants in the library prep
or something funky with the sequencers). In general, it's nice if during
any read filtering, there is an option to write out reads that are being
filtered out. I understand that this could probably slow things down, but
it's nice to have as a lab troubleshooting feature.

Thanks so much!
Julie

TomSmithCGAT:
OK, I'll try and get this sorted on a branch for you to test shortly.

TomSmithCGAT:
Hi Julie,

Sorry about the delay with this, got sidetracked!

Could you try installing from the `{TS}-OutputFilteredReads` branch. This includes two new options, `--filtered-out` and `filtered-out2` to write out filtered reads/read pairs to a separate file. I've checked that the number of filtered reads output matches the number of filtered reads. If you could check the output with your input is as expected that would be great.

Note that reads can be filtered out due to not matching the pattern regex (`--extract-method=regex`) and if the cell barcode doesn't match a whitelist (`--filter-cell-barcode`). As it stands, reads filtered for either reason will be written to the same file but the log will detail how many reads were filtered for either reason, e.g from the log file below, there were 10000 input reads, 4406 did not match the regex (and were therefore filtered out) and a further 170 reads had a cell barcode which could not be corrected to the whitelist:

```
2019-04-02 18:28:16,616 INFO Input Reads: 10000
2019-04-02 18:28:16,617 INFO regex matches read1: 5594
2019-04-02 18:28:16,617 INFO regex does not match read1: 4406
2019-04-02 18:28:16,617 INFO Reads output: 4070
[...]

PR #330, opened 2019-04-02
{ts} output filtered reads

TomSmithCGAT:
Do not merge until #328 is resolved. 

This branch adds options to write out reads filtered in `extract` to separate files for manual inspection. Calls to a `extract_methods.ExtractFilterAndUpdate()` object which result in a read not being fully processed generate a return value of `None`. This branch simply adds lines to write out the read(s) before the next call.


TomSmithCGAT:
Hi @IanSudbery. Would you mind taking a look at this PR. Idea is simply to allow the user to request that the filtered reads are output to separate fastqs for manual inspect. See #328 for the motivation

TomSmithCGAT:
@IanSudbery - Do you mind sanity checking this please

umi_tools/extract.py
```diff
@@ -285,7 +285,13 @@ def main(argv=None):
                               add_sam_options=False)
 
     if options.retain_umi and not options.extract_method == "regex":
-        raise ValueError("option --retain-umi only works with --extract-method=regex")
+        U.error("option --retain-umi only works with --extract-method=regex")
+
+    if (options.filtered_out and not options.extract_method == "regex" and
```
IanSudbery, line 290: Should this not be `options.filtered_out and (not options.extract_method == "regex" or not filter_cell_ barcodes)`

TomSmithCGAT: Ah, no.  `options.filtered_out` is only compatible with either extract using regex (`options.extract_method == "regex" ) *or* cell barcode filtering (`not filter_cell_ barcodes`), since these are the only times one wants to filter the reads in any way. Hence, error is raise if both are false.

I guess the simplest to write this would actually be `options.filtered_out and not (options.extract_method == "regex" or filter_cell_ barcodes)`

TomSmithCGAT:
Thanks for reviewing. I'll try and get to the bottom of the travis errors first before merging or else this will just error upon merge with master

PR #339, opened 2019-05-30
fix python2 ImportError

jocelyne8:
in python2: "ImportError: No module named builtins"
bytes and chr are built-in (with no import) in python2

### DIFF
--- a/umi_tools/Utilities.py
+++ b/umi_tools/Utilities.py
@@ -258,7 +258,10 @@ import tempfile
 import regex
 from umi_tools import __version__
 
-from builtins import bytes, chr
+try:
+    from builtins import bytes, chr # python3
+except:
+    pass # python2
 
 
 class DefaultOptions:


IanSudbery:
Thanks for your PR. We appreciate contributions from the community. 

While we are happy to accept PRs that maintain python2 compatibility, note that UMI-tools no longer supports python2 and we can't guarentee that this, and particularly future versions of UMI-tools or its dependencies will function properly in python2 or that simple fixes will make them compatible. That said, if you are willing to keep supplying the patches, I can't see any reason we wouldn't be willing to keep accepting them. 

For this particular PR there are just a couple of changes we'd like. We try to keep all of our code pythonic and PEP8 compliant. Your PR is failing our PEP8 test with an  E261 violation (which is "two spaces before an inline comment"). Also I noticed that you have an indiscriminate except that passes errors silently. Could you possible make it only catch ImportError? 

Thanks again.

TomSmithCGAT:
I'm going to close this PR. @jocelyne8 - Please do re-open if you still want to keep UMI-tools python 2 compatible and have made the changes Ian suggested above.

PR #365, opened 2019-08-22
Attempt to fix #349.

### LINKED ISSUE #349: How to use Pyinstaller on UMI-tools and make it reproducible?
Hi,

I was stumbling on an issue of using UMI-tools together with the generation of standalone (one-file) executable using  [pyinstaller](https://github.com/pyinstaller/pyinstaller). I use it to being able to distribute UMI-tools more easily to machines with same architecture but I noticed that the output was not reproducible anymore even if setting the `--random-seed` option. It seems that it is related to `PYTHONHASHSEED` since when using UMI-tools normally (e.g. from a virtualenv), setting this made it reproducible.

This is how I first installed UMI-tools (using python3.6 but I guess it is the same with other version of python3) and used it to run functional tests.
```
git clone https://github.com/CGATOxford/UMI-tools
cd UMI-tools/ && virtualenv -p python3.6 venv && . venv/bin/activate
pip install "numpy==1.16.4" "pandas==0.24.2" "future==0.17.1" "regex==2019.6.8" "scipy==1.3.0" "matplotlib==3.1.1" "pysam==0.15.2"
pip install .
export PYTHONHASHSEED=0 && nosetests -v tests/test_umi_tools.py
deactivate
```
=> All 53 test were successful

My procedure to build a one-file executable for UMI-tools with pystaller, including a [patch](https://drive.google.com/file/d/1Dfl_BtWx3fvivrg6Ak6Ga-_DZjKkQ6t3/view?usp=sharing) to circumvent an issue with Tcl/Tk libraries on OSX (pyinstaller/pyinstaller#3753) and also run the functional tests.
```
. venv/bin/activate && pip install "pyinstaller==3.4"
patch venv/lib/python3.6/site-packages/PyInstaller/hooks/hook-_tkinter.py pyinstaller_hook-_tkinter.patch
rm -rf build dist && pyinstaller --clean $(find venv/lib/python3.6/site-packages/umi_tools/ -maxdepth 1 \( -name "*.py" -o -name "*.c" \) | grep -v "__init__.py" | grep -v "umi_tools_exe.py" | xargs -n 1 -I {} echo "--add-data "{}":umi_tools") --hidden-import pysam --hidden-import pysam.libctabixproxies --hidden-import numpy --hidden-import numpy.matlib --hidden-import pandas  --hidden-import regex  --hidden-import scipy --hidden-import scipy._lib.messagestream --hidden-import matplotlib --hidden-import future.utils -F -n umi_tools venv/lib/python3.6/site-packages/umi_tools/umi_tools.py
deactivate
export PYTHONHASHSEED=0 && OLD_PATH=${PATH} && export PATH=${PATH}:${PWD}/dist && nosetests -v tests/test_umi_tools.py; export PATH=${OLD_PATH}
```
=>17 of 53 tests failed

To fix this I tried to set `PYTHONHASHSEED` explicitly in the modules (e.g. in `umi_tools/umi_tools.py`) or created a run-time hook that simply sets this environment variable but this also did not work. Maybe it is related to the way, I am creating the one-file executable but adding all python source files and the one c file as data was the only way I was successful in generating a functional one-file executable.

Just for completion, I am using Python 3.6.7 and umi_tools (e8c2b47) on Mac OSX.

Best regards,
Christian

christianbioinf:
Hi,

I know that this may not be a suitable solution for the community of umi_tools but I still wanted to provide my progress on resolving #349.

I found out that it is caused by the use of unordered structures like `dicts` (for python < 3.5) and `sets` (for all version of python3) since this order sometimes influences the results of dedup operations (e.g. the order in which clusters are being built). This "random" behaviour can be suppressed by setting the environment variable `PYTHONHASHSEED` before starting the interpreter. However, since I am freezing the application with `Pyinstaller`, it is currently not possible to set this before the interpreter inside the bootloader starts up (see pyinstaller/pyinstaller#3665).

As I am only interested in reproducibility with python3.5+, I simply replaced all relevant occurences of `sets` with `dicts` inside `network.py`. Note that I also tried to fix the recursive version of the breadth-first search algorithm (if someone is interested in this). For completeness, I updated the test result files and `nosetests` is now running fine even without setting `PYTHONHASHSEED`.

As I already mentioned, I am aware of the limitations of my fix (in the supported python versions) and the dependency on the order inside `dicts` which is currently only considered an implementation detail and not part of the spec yet. Also, my use case is rather specific but since more and more people and researchers are concerned about reproducibility, freezing python application (especially those with lot of python package dependencies) is important and hence it may be also interesting for a wider audience.

I am happy to take feedback or suggestions how I could improve it in order for it to be merged into the master branch.

Best,
Christian


IanSudbery:
Hi @christianbioinf The latest release should finally bring reproducible behavoir to UMI-tools without the need for custom fixes.

Issue #347, opened 2019-07-05
umi_tools dedup: Incorrect output with using --paired


christianbioinf:
Hej hej,

I noticed an issue with `umi_tools dedup` using the option `--paired` which in some cases reports the read alignments multiple times in the output. By that, I mean that one gets literally duplicate lines in the SAM/BAM format.

I have traced it down that it is in conjunction with the output writer used with `--paired`, namely the `TwoPassWriter` in `sam_methods.py`. It is supposed to output the second-in-pair read alignment that corresponds to reported first-in-pair read alignments. However, in case the first-in-pair and second-in-pair read alignments share the same reference name and reference start, it may report the first-in-pair read alignment twice. Also, it produces incorrect (or at least unexpected) output in case the same second-in-pair read alignment is referred to by multiple first-in-pair read alignments (which `bwa mem` does by default in case there are secondary read alignments).

I compiled a small testset of read alignments illustrating both cases and I am currently working on a fix that I will make available as pull request. But for those who are interested and also later for checking the fix by the pull request, please find the alignment file for download ([here](https://drive.google.com/file/d/1fLmR7TdtutPkvVhh8gJ65-kTp2IFdPLj/view?usp=sharing)). It contains the following read alignments (for those who do not want to download it).
```
TestRead_SECONDARY_ATAACCGGTTAC	163	chr18	13300195	60	127M	=	13300322	183	AGCAGGTGCCTCCTTACAGTTGGAGTCTGTCCAGGCACAGCAGAGAGCCAGGCCCGCAGAGGGGGCCGGCAGCCTCTTCCCACTCTCCTGCCCACCCCGCGCCTGTGCTCTGCCTCAGCCACAAGGC	HHHHHDAGHGHHHHHHHHHFGHHHHHHHHEFHGHGHHHHHHHHGFFEHGHHEEFGEGGGGFFGGGGGECEGCDHHHHGHHHEFFHBGFFFHHFEGGGGGGGGGFEGGFFHHHGH0GHBGGHHFEFGH	NM:i:0	MD:Z:127	AS:i:127XS:i:0	RG:Z:HD200	MQ:i:60	MC:Z:44S56M51S
[...]

PR #366, opened 2019-08-23
Fix for incorrect output of dedup with --paired (#347) in case of duplicated lines

christianbioinf:
Due to the feedback from @IanSudbery, I have now just created a pull request for the trivial and easy fix for having multiple identical read1 alignments in the case the read2 alignment is missing by extending one condition.

Best,
Christian


IanSudbery:
Great! Thanks

Issue #261, opened 2018-06-29
which read is kept upon deduplication


retrogenomics:
Hi,

I'm wondering what is the rationale for keeping a specific read in a pile of duplicates. Sometimes the read kept is short or has indels/sequencing errors. In Picard tools, there is an option to keep the longest, or the one with the best quality score. Is there a similar option in UMI-tools?

Thanks in advance for your help

IanSudbery:
For each positions we apply our dedup algorithm and select for each group the barcode that could not have arisen as a PCR error from the others.  This is generally the UMI with the highest read count. Within the reads with the selected UMI, we choose the read that has the highest mapping quality. If two reads have the same mapping quality, we select the one with the lowest number of multimapping hits. If reads have the same mapping quality, and the same number of multimapping hits, we choose the read at random. 


Thus you shouldn't be getting a low quality read selected when one with a higher quality and the same UMI is available.

retrogenomics:
I was unprecise: from what you described, your selection procedure does not seem to take into account read _sequencing quality_ (only mapping quality). In other words, if 2 reads have the same high score _mapping quality_ (i.e. unique mappers), one being long and with good base scores, the other short with errors, it will select randomly among these, too, right?

TomSmithCGAT:
Yes, that's correct. Honestly, this is because during the development of UMI-tools, the major techniques we had in mind (iCLIP & scRNA-Seq) aren't concerned about the precise sequence, just the alignment coordinates, so it didn't really matter which final read was retained

However, if you would like to make sure you retain the read with the highest read quality, you can run `umi_tools group` (should drop in for `dedup` will all the same options) and then parse the BAM file to select the read from each group with the highest sequence quality (each read will be given a `UG` tag which denotes the group of duplicate reads from which it originates).

retrogenomics:
Thanks for the clarification.

PR #378, opened 2019-10-31
Clarified Selecting the representative read documentation

mikej888:
A user queried the use of quality in which read is kept upon
deduplication (#261)

> your selection procedure does not seem to take into
> account read sequencing quality (only mapping quality).
>
> In other words, if 2 reads have the same high score
> mapping quality (i.e. unique mappers), one being long and
> with good base scores, the other short with errors, it will
> select randomly among these, too, right?

The response was:

> Yes, that's correct.

Updated umi-tools/dedup.py "Selecting the representative read" comment
section to clarify that the read quality is not used.

### DIFF
--- a/umi_tools/dedup.py
+++ b/umi_tools/dedup.py
@@ -23,7 +23,10 @@ will be retained from a group of duplicated reads:
 1. The read with the lowest number of mapping coordinates (see
 ``--multimapping-detection-method`` option)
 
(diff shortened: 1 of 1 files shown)


TomSmithCGAT:
@mikej888 - Thanks for adding this clarification

mikej888:
Thanks!

PR #385, opened 2019-12-06
avoids keyError when generating random reads

TomSmithCGAT:
@IanSudbery - You happy with this? The KeyError is already picked up also during the read parsing and the user will be informed at this point how many missing cell tags there were and an example read given.

I'll update the tests to cover this too before merging.


TomSmithCGAT:
Should have said that this relates to  @ZHUwj0's error described in #281.

I've updated the tests to capture this now and bumped the version number.

On the basis that this is only a very straightforward debug, I'm going to go ahead and merge this into master immediately and issue version `1.0.1`.

PR #399, opened 2020-02-24
Extended regex.md 'Regex (regular expression) mode' with examples

mikej888:
Added two example sections:

* Example: extracting UMIs
* Example: extracting barcodes and UMIs

These have example regexes and FASTQ fragements and show
how the regexes match the FASTQ reads and the output FASTQ
files that would be produced.


IanSudbery:
Thanks, this looks great.

mikej888:
@IanSudbery Thanks!

Issue #395, opened 2020-02-06
Docs request: more detailed description of outputs with dedup --output-stats option


ijhoskins:
I took a look at the manual regarding the three output files generated with this option, but looking at the files some things are still not clear to me. The manual describes a little more in detail what the "_stats_edit_distance.tsv" columns mean, but the others are less clear.

For example, the name of the "_stats_per_umi_per_position.tsv" file suggests we should see read counts pre- and post- dedup for each [position x UMI] combination. But there are no columns that report coordinate position or UMI identity.

Similarly, for the "_stats_per_umi_per.tsv" file, it is not entirely clear the difference between the "times_observed" and "total_counts". Is the difference that the former is counts for the originating UMI whereas the latter is the total counts for the originating UMI and all variants within the cluster?

Any elaboration on the nature of these values would be much appreciated!

TomSmithCGAT:
Hi @ijhoskins. Apologies for the delay responding to your query.

I agree the name for these files is a bit confusing. The `_stats_per_umi_per_position.tsv` file simply tabulates the counts for unique combinations of UMI and position. E.g if prior to deduplication, we have two positions in the BAM (POSa, POSb), at POSa we have observed 2*UMIa, 1*UMIb and at POSb: 1*UMIc, 3*UMId, then the stats file is populated thus

| counts  | instances_pre|
| ------------- | ------------- |
| 1  | 2  |
| 2| 1 |
| 3| 1 |

If post deduplication, UMIb is grouped with UMIa such that POSa: 3*UMIa, then the `instances_post` column is populated thus:

TomSmithCGAT:
Hi @ijhoskins. Apologies for the delay responding to your query.

I agree the name for these files is a bit confusing. The `_stats_per_umi_per_position.tsv` file simply tabulates the counts for unique combinations of UMI and position. E.g if prior to deduplication, we have two positions in the BAM (POSa, POSb), at POSa we have observed 2*UMIa, 1*UMIb and at POSb: 1*UMIc, 3*UMId, then the stats file is populated thus

| counts  | instances_pre|
| ------------- | ------------- |
| 1  | 2  |
| 2| 1 |
| 3| 1 |

If post deduplication, UMIb is grouped with UMIa such that POSa: 3*UMIa, then the `instances_post` column is populated thus:

| counts  | instances_pre|instances_post|
| ------------- | ------------- |------------- |
| 1  | 2  |1|
| 2| 1 |0|
[...]

Issue #381, opened 2019-11-16
I need a bam file that contains an error corrected molecular barcodes as a TAG named UB or XM. 


sarajimenez:
Hello, 

I'm sorry, I have a very basic question ...

I need a bam file that contains an error corrected molecular barcodes as a TAG named UB or XM. Specifically for running velocyto. 

I've been struggling to try to add this modifying several options with the command _count_.  

Right now my bam file looks like this:

`K00201:191:HNVFNBBXX:1:1225:27600:17034_GTGAGTGA 272 chr1 3329451 1 48M12S * 0 0 AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAATCACTCACAGCACGGTTA A<-FF<FAAA7-7--J<F-JFFJJJJJJJJJJJJJJJJJJJJJJJJJJJJJJJJJFFFAA NH:i:3 HI:i:3 NM:i:1 MD:Z:45G2 AS:i:45 nM:i:1 jM:B:c,-1 jI:B:i,-1`

And I need somethis like this: 

`K00201:191:HNVFNBBXX:1:1225:27600:17034_GTGAGTGA        272     chr1    3329451 1       48M12S  *       0       0     AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAATCACTCACAGCACGGTTA     A<-FF<FAAA7-7--J<F-JFFJJJJJJJJJJJJJJJJJJJJJJJJJJJJJJJJJFFFAA   NH:i:3  HI:i:3  NM:i:1  MD:Z:45G2       AS:i:45 nM:i:1  jM:B:c,-1       jI:B:i,-1       UG:i:0  XM:Z:GTGAGTGA`

I'd highly appreciate your help!

Have a nice weekend.

IanSudbery:
If you use `group`, with `--out-bam` rather than `count`, then the error
corrected output will be added to the read as tag BX. See
https://umi-tools.readthedocs.io/en/latest/reference/group.html

Now I can think of several ways of renaming this tag `BX`, but they all
feel a bit clunky. Here is how I would do it, using a python program and
pysam:

```python
import pysam
bamfile = pysam.AlignmentFile("mybam.bam")
with pysam.AlignmentFile("mybam_processed.bam", "wb", template=bamfile) as
outfile:
     for read in bamfile.fetch(until_eof=True):
          umi = read.get_tag("BX")
          read.set_tag("UM", umi)
[...]

PR #401, opened 2020-02-26
updates documenation for stats option

TomSmithCGAT:
Explanation of stats files was insufficient

See #395 for motivation.

### DIFF
--- a/umi_tools/dedup.py
+++ b/umi_tools/dedup.py
@@ -63,15 +63,50 @@ Each unique UMI sequence may be observed [0-many] times at multiple
 positions in the BAM. The following files report the distribution for
 the frequencies of each UMI.
 
-[PREFIX_stats_per_umi_per.tsv
-  For each UMI, the positions was it observed at (times_observed), the
-  sum of freuqencies across all positions (total_counts) and the
-  median for the distibution of frequencies (median_counts). These are
-  all presented pre- and post-deduplication
-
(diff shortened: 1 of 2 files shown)


TomSmithCGAT:
Don't merge. I'll use this branch to update documention for `--output-bam` and `--umi-group-tag` options too (see #381)

ijhoskins:
Hi @TomSmithCGAT some suggestions:

I think you should use your original descriptions you commented in https://github.com/CGATOxford/UMI-tools/issues/395

Currently [PREFIX_stats_per_umi_per.tsv is ambiguous. Here you mean [PREFIX] _stats_per_umi.tsv.

For this one I like the descriptions:
times_observed = how many positions the UMI was observed at
total_counts = the total number of times the UMI was observed across all positions
median_counts = the median for the distribution of how often the UMI was observed at each position (excluding zeros)

The term frequencies is a little misleading here as they are reported as counts. Additionally, there are also some typos in this description: "For each UMI, the positions was it observed at".

For _stats_per_umi_per_position.tsv I think again the example tables you originally sent are useful. Also, you might change the header as such:
counts -> UMI_duplicates
instances_pre - > positions_pre
instances_post - > positions_post

I would also avoid the term frequencies here as well as they are reported as counts. You might also state that both instances_pre and instances_post <= targeted_positions although counts is only bounded by depth of sequencing.

Thanks again for clarifying and updating the docs!

Issue #393, opened 2020-01-16
 Option to extract from either read (--either-read) is not currently compatible with cell barcode extraction


bioarpit1:
Hi I am trying to use barcode extraction for my STARR seq reads where my barcodes are on 5" end of read2, but some times read1 also contains the barcode on 3" end. Here is the extraction log. 
# UMI-tools version: 1.0.0
# output generated by extract --extract-method=string --either-read --bc-pattern=GGTCAGAATGATNNNNN --bc-pattern2=NNNNNATCATTCTGACC --either-read-resolve=quality --stdin=dTC1_R1.fastq.gz --stdout=dTC1_R1_extracted.fastq.gz --read2-in=dTC1_R2.fastq.gz --read2-out=dTC1_R1_extracted.fastq.gz -L extract.log
# job started at Wed Jan 15 15:29:51 2020 on h001.grid.gs.washington.edu -- 452fa656-6b3e-48fe-96cb-a857a2b1484c
# pid: 176561, system: Linux 2.6.32-696.30.1.el6.x86_64 #1 SMP Tue May 22 03:28:18 UTC 2018 x86_64
# blacklist                               : None
# compresslevel                           : 6
# either_read                             : True
# either_read_resolve                     : quality
# error_correct_cell                      : False
# extract_method                          : string
# filter_cell_barcode                     : None
# filter_cell_barcodes                    : False
# log2stderr                              : False
# loglevel                                : 1
# pattern                                 : GGTCAGAATGATNNNNN
# pattern2                                : NNNNNATCATTCTGACC
# prime3                                  : None
# quality_encoding                        : None
# quality_filter_mask                     : None
# quality_filter_threshold                : None
# random_seed                             : None
# read2_in                                : dTC1_R2.fastq.gz
# read2_out                               : dTC1_R1_extracted.fastq.gz
# read2_stdout                            : False
# reads_subset                            : None
# reconcile                               : False
# retain_umi                              : None
# short_help                              : None
# stderr                                  : <_io.TextIOWrapper name='<stderr>' mode='w' encoding='UTF-8'>
# stdin                                   : <_io.TextIOWrapper name='dTC1_R1.fastq.gz' encoding='ascii'>
# stdlog                                  : <_io.TextIOWrapper name='extract.log' mode='a' encoding='UTF-8'>
# stdout                                  : <_io.TextIOWrapper name='dTC1_R1_extracted.fastq.gz' encoding='ascii'>
# timeit_file                             : None
[...]

PR #402, opened 2020-02-26
adds checks that barcode string contains only NCX characters

TomSmithCGAT:
This introduces a check that barcode strings do not contain non-NCX characters

See #393 for motivation.


TomSmithCGAT:
I've run a manual check for expected behaviour but we don't cover errors in our testing so would appreciate a sanity check here.

Issue #411, opened 2020-04-30
--error-correct-cell behaviour


mbahin:
Hi,

Thanks for the tool.
I have a problem with --error-correct-cell, I don't know if I misunderstood something.

So first I do a whitelist:
`umi_tools whitelist --extract-method=regex --bc-pattern='(?P<discard_1>TTCG){s<=1}(?P<cell_1>.{15,17})(?P<discard_2>TGCTTACGCTACGGAACGA){s<=3}(?P<umi_1>.{9})' --stdin=input.fastq -S BBC.txt`
Which give me in particular this line in the whitelist output file:
CTGTTGATCACCCGTA	CTGTTGATCACCCGTAT
And then I do an extract:
`umi_tools extract --extract-method=regex --bc-pattern='(?P<discard_1>TTCG){s<=1}(?P<cell_1>.{15,17})(?P<discard_2>TGCTTACGCTACGGAACGA){s<=3}(?P<umi_1>.{9})' --whitelist BBC.txt --error-correct-cell --stdin=input.fastq -S input.BBC.fastq`
And in the output FASTQ file, I have this read header:
@M05218:191:000000000-D7R5H:1:1102:14341:19429_CTGTTGATCACCCGTAT_CCTCAAACG 1:N:0:2
where I was expecting the BC to be corrected to "CTGTTGATCACCCGTA" but it actually has the uncorrect form (that was actually found in the read).
Is it the expected behaviour? Did I misunderstand something?

Cheers,
Mathieu

IanSudbery:
I think you need `--filter-cell-barcode` as well. Sorry, that isn't well documented.

mbahin:
Yes it works with this additional option, thanks! :)

mbahin:
Actually I don't understand why "--filter-cell-barcode" and "whitelist" are not one unique option?

TomSmithCGAT:
There is indeed a redundancy between `--whitelist` and `filter-cell-barcode`, where the former is a path to a whitelist file and the later a switch to filter against this. `error-correct-cell` is a separate option however since one might wish to only retain cells than perfectly match the whitelist

Issue #424, opened 2020-07-23
Silent out of bounds failure 


pati-ni:
`umi_tools extract` fails with an out of bounds exception when read is shorter than `--bc-pattern`

An option that ignores such reads would be useful in processing such files.

Currently on illumina reads, the workaround is to re export fastq files with `bcl2fastq` which requires extra software along with the bcl files.

TomSmithCGAT:
Hi @pati-ni - Sorry for the very delayed reply here.

Can you confirm how the read length comes to be shorter than the bc-pattern in the first place? Is this due to trimming of the reads before UMI extraction and if so, is the trimming done within bcl2fastq and hence the need to re-export with `bcl2fastq`?

If this is the case, I would argue that the input to `umi_tools extract` is pretty poorly generated in the first place. I would prefer to see this resolved by exporting with `bcl2fastq` with `-minimum-trimmed-read-length` set to a more suitable value, or else read trimming performed post UMI extraction. I think handling this within `umi_tools extract` is probably getting outside of the functionality it should cover. Thoughts @IanSudbery?

IanSudbery:
We should probably at least have a better error message.

TomSmithCGAT:
Yep. Very good point. Let's leave this open until that is addressed.

TomSmithCGAT:
This was resolved with PR #433

PR #433, opened 2020-09-14
{ts} documentation update

TomSmithCGAT:
When filtering against a whitelist with `extract`, no longer need to supply `--filter-cell-barcode` and `--whitelist`, just the later. Note that former option remains but is hidden from user in `--help` text and will raise an `U.info` to inform the user this option is no longer required. See #411 for motivation

Also adding a specific doc page to walk through the network-based deduplication. This is following an email conversation with a user who pointed out some inadequacies/inconsistencies in the way the methods were described across the original CGAT blog post/GR publication/command-line documentation.


TomSmithCGAT:
Also now checks that string patterns are not longer than read length and if so, outputs more understandable error (see #424 for motivation)

TomSmithCGAT:
Nothing else in the open issues for documentation/minor debugs etc so this branch is ready for review

IanSudbery (approved):

Issue #428, opened 2020-09-07
sam_methods.get_bundle code


mitrinh1:
Hi all, thanks so much for developing this great tool!

I have some questions regarding the code of sam_methods.get_bundles class:

- line 202: self.start=0 never gets updated?
- line 481: 'start' returned by get_read_position() is not used anywhere else? What is the purpose of this? 
- Additionally, I am not entirely clear on the differences between 'start' and 'pos' generated by get_read_position(). It seems that soft_clipping is taken into account in calculating 'start' in forward reads, but not for reversed reads?
 
Also, am I correct in saying that the umi_tools._dedup_umi.edit_distance method calculates hamming distance?

Thank you in advance for your kind help! It is much appreciated!

TomSmithCGAT:
Thanks for noting the odd behaviour with `self.start`  and `start` @mitrinh1. Having looked into this, it turns out this was an oversight on my behalf when the code was refactored to deal with single cell RNA-Seq data. Since `self.start` is not updated, this also means that bundles don't get yielded until the contig changes which is not the intended behaviour. Whilst this behaviour ensures that no reads are missed, this is not the intended default behaviour, which is to yield bundles when the current read position is 1000 bp downstream of the bundle. In essence, I've done the equivalent hard-coding the `--buffer-whole-contig` option to be on, which will have a negative impact on memory usage. 

I'll run a few tests to see how big the impact might be and patch this promtly!

TomSmithCGAT:
On your other points, you're correct that `pos` takes account of soft clipping and that `edit_distance` calculates the hamming distance

TomSmithCGAT:
For completeness, here is the amount of memory required to process a test input (~2.5M paired end) using the 'current' code and the 'updated' code where `self.start` updates properly and bundles are yielded as intended. With the intended behaviour, there's a small but clear reduction in memory use, as expected.





| code    | buffer_whole_contig | mem (kb) |
|---------|---------------------|----------|
| Current | false               | 1782296  |
| Current | true                | 1754400  |
| Updated | false               | 1381940  |
| Updated | true                | 1754756  |

mitrinh1:
Thank you so much for this! It is much appreciated!

mitrinh1:
[...]

PR #434, opened 2020-09-14
Enables reads to be processed before end of contig

TomSmithCGAT:
As noted in #428, `get_bundle.start` was never updated. As such `get_bundles.check_output` only returned `do_output=True` at the end of a contig which is not the intended behaviour. This PR fixes this.

Note that this also changes the output order for some test files. For each of the test files modified herein, I have manually checked (`diff <(sort new_test_output) <(sort old_test_output)`) to make sure the only difference is the order of the output.


IanSudbery (commented):
looks good, but unfortunately look like you'll need to update the test files again :(

PR #443, opened 2020-10-28
Fix bug with unmapped paired end reads

Daniel-Liu-c0deb0t:
It seems that for paired end reads where the second read of each pair is unmapped, they are not discarded when the option for handling unmapped reads is to discard them. I'm not entirely sure whether this is intended behavior, but it likely isn't and the code has a typo.


IanSudbery:
I'm 90% sure you are right, but better wait for @TomSmithCGAT to chime in as he wrote that bit.

TomSmithCGAT:
Hi @Daniel-Liu-c0deb0t. Thanks for catching this. You're absolutely right that this isn't the intended behaviour.

The travis fail is because the test files covering this option need updating. Would you like me to update them on your branch so the PR can be merged?

Daniel-Liu-c0deb0t:
Go ahead and push to my branch. You can merge when you have updated the tests.

PR #452, opened 2021-01-22
fix read1 error in writing filtered result

yech1990:


IanSudbery:
Is it possible to link this to an issue describing the problem you are trying to fix? 

As far as I can tell, your change at line 446 looks reasonable, but the change at 427 looks like it means that no read1 will ever be output unless output of filtered reads is set to true. I assume this is why the PR test is failing.

yech1990:
clsoe #453

IanSudbery:
Thanks for contributing, we appreciate it.

PR #464, opened 2021-03-18
Fixing --whitelist command-line option 

jonn-smith:
- added missing reference to `options` object.


TomSmithCGAT:
Thank you for this PR and sorry for completely missing it. Ian separately issued an identical PR #510 which I've just merged, or else I would merge this one!

PR #506, opened 2022-01-18
Fix "umi_tools command" prints stack trace

epruesse:
Also:
- minor refactor
- pass return value from command through to sys.exit()
- pretty error if the command is not known

(Didn't test this PR myself yet, just trying to fix my "first launch" experience of getting a stack trace)


IanSudbery:
Looks good, but I will need to test it.

Issue #536, opened 2022-06-02
Conda issue: IndexError: list index out of range


CWYuan08:
Hi, umi_tools has always been working for me, but suddenly when I tried to put together a pipeline, it came to this error, for any command I use after umi_tools, (e.g. umi_tools dedup..):

Traceback (most recent call last):
  File "/miniconda3/envs/umitools/bin/umi_tools", line 8, in <module>
    sys.exit(main())
  File "/miniconda3/envs/umitools/lib/python3.7/site-packages/umi_tools/umi_tools.py", line 53, in main
    elif argv[2] in ["--help", "-h", "--help-extended"]:
IndexError: list index out of range

However, when I just type umi_tools, there isn't any error.

Do you know how I could resolve this? 

Thank you very much!

CWYuan08:
Hi @IanSudbery, sorry to bother again, I'm still getting the same issue, what should I do to resolve this..? Thank you very much!

Best,
CW

IanSudbery:
Hi,

You are getting this error because you are executing `umi_tools dedup` etc, without any other options, and `dedup` etc requirement than just the name of the tool. We have edited the code so that the help should be output in this situation (if you pull the latest version of the master branch), but even with the last released version, you can avoid this by typing `umi_tools dedup --help` in order to get the help.

CWYuan08:
Dear @IanSudbery,

thank you. I see, it worked!

Best regards,
CW

PR #537, opened 2022-06-05
Altered if clasuses around selection of commands in umi_tools.py

IanSudbery:
Fixes (or at least starts to fix) #536 

@TomSmithCGAT Bit confused as to what is going on in the umi_tools module. Specifically what the original purpose of the if clause at lines 51-52 was.

I've also made each tool print its usage message if no options are passed at all. Don't know if this was the right was to do this, or if I should have done something more central (although that would become a problem if there was a tool that could be run without options. Which currently there isn't).


umi_tools/Utilities.py
```diff
@@ -1030,7 +1030,7 @@ def Start(parser=None,
 
     if return_parser:
         return parser
-
+
```
TomSmithCGAT, line 1033: Is that just a tab added?

umi_tools/count_tab.py
```diff
@@ -86,12 +86,17 @@ def main(argv=None):
 
     if argv is None:
         argv = sys.argv
-
+
```
TomSmithCGAT, line 89: Another tab to remove?

umi_tools/count_tab.py
```diff
@@ -86,12 +86,17 @@ def main(argv=None):
...
     if argv is None:
         argv = sys.argv
-
+        
     # setup command line parser
     parser = U.OptionParser(version="%prog version: $Id$",
                             usage=usage,
                             description=globals()["__doc__"])
 
+    if len(argv) == 1:
```
TomSmithCGAT, line 95: Happy with this to catch use of tool with no options

umi_tools/umi_tools.py
```diff
@@ -50,12 +48,22 @@ def main(argv=None):
...
         print("UMI-Tools: Version %s" % __version__)
 
     command = argv[1]
 
-    module = importlib.import_module("umi_tools." + command, "umi_tools")
+    try:
+        module = importlib.import_module("umi_tools." + command, "umi_tools")
+    except ImportError:
+        print("Command %s not recognosied" % command)
+        print()
```
TomSmithCGAT, line 60: ?

umi_tools/umi_tools.py
```diff
@@ -50,12 +48,22 @@ def main(argv=None):
 
         return
 
-    elif argv[2] in ["--help", "-h", "--help-extended"]:
+    elif len(argv) > 2 and  argv[2] in ["--help", "-h", "--help-extended"]:
```
TomSmithCGAT, line 51: I think this elif can be removed. The tools all have -h/--help/--help-extended options so no need to catch anything here

TomSmithCGAT:
Hey @IanSudbery. Assuming you're referring to and the line numbers are from this PR?
https://github.com/CGATOxford/UMI-tools/blob/0c72fa99c518088aa4505ee16721f85ac42eb0c2/umi_tools/umi_tools.py#L53-L54

I'm not sure what was going on with that elif either. It looks like I added it when we split into multiple tools to print help when a tool is specified and then you changed the output to be the version number (0101494bf236a485cbb14a55255113374df3c9a9 ). Regardless, each tool command line has help options added so no need to catch it in `umi_tools.py`.

I'm fine with having the same code in each tool to output help if no options are given. I agree it would be best done centrally though if you can be bothered recoding. I guess the easiest way to do it would be to catch instances where `len(sys.args)==2` after the checks for version/help options, print the line you have about needing to add required options and then add `--help` to the sys.args and call the tool module main. I can't imagine us adding a tool with no help in the future since our base arg parser adds `--help` anyway

TomSmithCGAT:
Can't see how to comment on lines that haven't been updated in the PR but L46 in `umi_tools.py`:

`elif len(argv) == 1 or argv[1] == "--version" or argv[1] == "-v":`

That `len(argv) == 1` can go, since that case is met by the `if` and causes the help to be printed.

Minor thing.

umi_tools/count_tab.py
TomSmithCGAT: There are a few of these in the PR. One in each tool it looks like. Unless I'm missunderstanding the github diff markup.

umi_tools/umi_tools.py
TomSmithCGAT: What's the `print()` for?

umi_tools/count_tab.py
IanSudbery: I don't think you are misunderstanding.

umi_tools/umi_tools.py
IanSudbery: Lazy way to add a blank line 😬

umi_tools/umi_tools.py
IanSudbery: My guess is that this was added because the standard `--help` within each tool doesn't output the version number?

IanSudbery:
>  can't imagine us adding a tool with no help in the future since our base arg parser adds --help anyway

Thats not what I meant. I guess I was thinking about a tool that had no required arguements. For example, you could imagine that if `extract` had a default `--pattern` it could be run:

```
cat my_reads.fastq | umi_tools extract > my_reads.extracted.fastq
```

Although of course it doesn't, and I think all our tools require options of some sort (although come to think of it the new `prepare_for_rsem` or whatever we are going to call it might not). 

In the above `len(argv) == 1`, but there is nothing wrong with it. Although perhaps this is just premature optimisation on my part. 

I'll clean up the rough tabs.

umi_tools/umi_tools.py
TomSmithCGAT: Ah, maybe. 

On a related note, the `--version` options for each tool also don't properly return the version. Should probably just remove that option from the tools and make it available via `umi_tools --version` only

IanSudbery:
Whitespace should be tidied up.

TomSmithCGAT (approved):

TomSmithCGAT:
Noticed in passing that a lot of the changes to umi_tools.py are also in #506. Should we first merge that one so that the contributions include @[epruesse]?

IanSudbery:
yes
Ian Sudbery
Senior Lecturer in Bioinformatics,
Sheffield Institute for Nucleic Acids,
School of Biosciences,
The University of Sheffield.

web: www.sudlab.co.uk
Tel: 0114 222 2738
Twitter: IanSudbery
Show Calendar Availability
<https://calendar.google.com/calendar/u/0?cid=aS5zdWRiZXJ5QHNoZWZmaWVsZC5hYy51aw>

TomSmithCGAT:
@IanSudbery - This is OK to merge, right?

PR #543, opened 2022-07-07
Speed up writing mates

### LINKED ISSUE #539: Problems with UMI dedup and time when aligned against transcriptome
Hello,

I'm running a pipeline that uses UMI dedup. However, we see a that when deduplicating the alignment against the transcriptome we see that it takes a huge amount of time (it can go up to more than 24h). We did not see this problem when aligning against the genome. We also have seen that the program reads a huge amount of data (around 4Tb), which we also didn't see when deduplicating the alignment against the genome, when the program read around 30Gb of data.

In our case we have a set of 12 UMIs at the beginning of the first read (we extracted the UMI using this: --bc-pattern=NNNNNNNNNNNN), the Fastq files have 15M reads, paired-end, we align with STAR and the resulting bam files weight around 400Mb (we align around 50%).

We did the libraries with very low concentrations of RNA and we did multiple PCR cycles, so we would expect the samples to be very duplicated, which I suppose could increase the time.

Which parameters could I change to improve the speed of the process? 

Thank you very much,
Lluc

IanSudbery:
Fixes #539.

When `TwoPassPairWriter` reaches the end of a contig, it calls `write_mates`, which reopens that contig from the start and scans through for remaining mates. To do this, `write_mates` uses `pysam.AlignmentFile.fetch(...., multiple_iterators=True)`. As the file uses is the same filehandle being used by `bundle_iterator`, then `multiple_iterators=True` ensures that the position in the file is not lost in this operation. 

`multiple_iterators=True` imposes some overhead. With a reasonable number of contigs this is not a problem, as the overhead is small compared to the cost of the scan. However, when alignment is done to the transcriptome, `write_mates` is called 100s of thousands of times, and for some reason, `fetch` calls the `__init__` of `psyam.RowIteratorRegion` 4 times for each call to `fetch`. This causes a serious slow down, such that adding `--paired` to the commandline slows the processing for an example file down from a couple of minutes to five hours. 

This PR changes `TwoPassPairWriter` so that it's `__init__` opens a second file handle to the input file. This allows it to drop the requirement to use `multiple_iterators=True` and returns the performance to near that of the performance without `--paired`. 

As far as I can tell, this does not change the output (i.e. the two handles act independently). I have tested this both on the test files and on an example transcriptome alignment provided in #539. 

Time to run is reduced from 5 hours to 200 seconds.

### DIFF
--- a/umi_tools/sam_methods.py
+++ b/umi_tools/sam_methods.py
@@ -12,6 +12,9 @@ import collections
 import re
 import random
 from functools import partial
+
+import pysam
+
 import umi_tools.Utilities as U
 
 
@@ -572,6 +575,7 @@ class TwoPassPairWriter:
 
     def __init__(self, infile, outfile, tags=False):
         self.infile = infile
+        self.mate_file = pysam.AlignmentFile(infile.filename)
         self.outfile = outfile
         self.read1s = set()
         self.chrom = None
@@ -601,7 +605,7 @@ class TwoPassPairWriter:
             U.debug("Dumping %i mates for contig %s" % (
                 len(self.read1s), self.chrom))
 
-        for read in self.infile.fetch(reference=self.chrom, multiple_iterators=True):
+        for read in self.mate_file.fetch(reference=self.chrom, multiple_iterators=False):
             if any((read.is_unmapped, read.mate_is_unmapped, read.is_read1)):
                 continue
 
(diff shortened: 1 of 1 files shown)


TomSmithCGAT (approved):

TomSmithCGAT:
Wow, that's a pathologically bad performance UMI-tools has had for transcriptome alignments 😱 Good catch!

PR #546, opened 2022-07-11
Ts add test action

TomSmithCGAT:
Adds github actions to perform testing. 

Still relying on the same bespoke test script we repurposed from cgat which utilises `nose`. Note that `nose` is no longer supported and cannot be used to test python 3.10 (https://github.com/nose-devs/nose/issues/1099). I tried a few of the suggestions in that issue to resolve the issue but nothing worked straight out of the box. 

For now, I've dropped testing for `python 3.10` in favour of getting the CI testing back up and running. In the long run, it seems like we should be switching to `pytest`.


IanSudbery (approved):
Looks fine, although I'm a bit confused about python versions. The yml suggests python 3.7, 3.8 and 3.9. Does that mean it will test these versions? I remember you saying it didn't work on 3.10, but have we dropped support for 3.6?

PR #548, opened 2022-07-23
Add option to specify UMI separator in extract

opplatek:
Add option to specify UMI separator in umi_tools extract. The default stays as it is now ("_"). Allows users to change it to other strings.


TomSmithCGAT:
See comment in #549

PR #549, opened 2022-07-23
Add option to specify UMI separator in extract 2

opplatek:
Add option to specify UMI separator in umi_tools extract. The default stays as it is now ("_"). Allows users to change it to other strings. 2

### DIFF
--- a/umi_tools/extract_methods.py
+++ b/umi_tools/extract_methods.py
@@ -8,16 +8,16 @@ import umi_tools.Utilities as U
 from umi_tools.umi_methods import RANGES
 
 
-def addBarcodesToIdentifier(read, UMI, cell):
+def addBarcodesToIdentifier(read, UMI, cell, umi_separator):
     '''extract the identifier from a read and append the UMI and
     cell barcode before the first space'''
 
     read_id = read.identifier.split(" ")
 
     if cell == "":
-        read_id[0] = read_id[0] + "_" + UMI
+        read_id[0] = read_id[0] + umi_separator + UMI 
     else:
-        read_id[0] = read_id[0] + "_" + cell + "_" + UMI
+        read_id[0] = read_id[0] + umi_separator + cell + umi_separator + UMI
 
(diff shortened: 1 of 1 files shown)


IanSudbery:
Hi @TomSmithCGAT can we set up the CI to run on PRs as well as commits? 

@opplatek This looks good, but I'll need to run our tests.

TomSmithCGAT:
I just need to finish off #546 (add py 3.5) and then merge and it should run on all pushes

https://github.com/CGATOxford/UMI-tools/blob/cd6cfa72adabfe8a8e8b1ea97bd4653ffa9cfec0/.github/workflows/nosetests.yml#L3

I've blocked out tomorrow for just UMI-tools so I'll do this first thing and then try and resolve the deterministic issues

TomSmithCGAT:
Merged #546 and set to run CI on PRs and pushes. Need to work out how to get it to run on an existing PRs now though...
If all else fails, I guess it's a re-issue PR or add new commit with e.g empty line

TomSmithCGAT:
Hmm... so possible to set the workflow to allow a manual dispatch, but that only covers branches/tags on this repo, not PRs.

@opplatek, sorry to be a pain, but could you please issue a new PR and that should trigger the testing now. Thank you 🙏

TomSmithCGAT:
💡 Just need to close and re-open and github treats it as a new PR 😁

TomSmithCGAT:
@opplatek - Am I right in understanding that the two PRs from your `patch-1` and `patch-2` branches both need to be merged? I don't think the testing is appropriate for the separate PRs, since `patch-1` fails the tests because of an issue resolved in `patch-2` and `patch-2` passes the tests because it doesn't have any effect without the changes in `patch-1`. 

Could you please merge the two patches and issue a new PR 🙏 Thank you!

opplatek:
Hi @TomSmithCGAT . Sorry for the confusion. The new push of the changes is at #551.

PR #570, opened 2023-01-03
Handle `--per-contig` grouping with no gene tags

### LINKED ISSUE #568: group with `--per-contig` unnecessarily requires gene tags
The `--per-contig` flag requires the `--per-gene` flag, and if outputting a TSV, ends up in this branch of the code:

https://github.com/CGATOxford/UMI-tools/blob/a5b348993cd220c32a524cf77155c8e42c89a3e1/umi_tools/group.py#L287-L291

This results in an error if there are no gene tags. I believe this is unnecessary because the `--per-contig` flag already entails that `gene` should be equated to the `contig`. Namely, we could instead handle the case like:

```python
  if options.tsv:
    if options.per_gene:
      if options.per_contig:
        gene = read.reference_name
      else:
        gene = read.get_tag(gene_tag)
    else:
      gene = "NA"
```

mfansler:
Resolves #568. Please let me know if you would like a test added.

### DIFF
--- a/umi_tools/group.py
+++ b/umi_tools/group.py
@@ -286,7 +286,10 @@ def main(argv=None):
 
                     if options.tsv:
                         if options.per_gene:
-                            gene = read.get_tag(gene_tag)
+                            if options.per_contig:
+                                gene = read.reference_name
+                            else:
+                                gene = read.get_tag(gene_tag)
                         else:
                             gene = "NA"
                         mapping_outfile.write("%s\n" % "\t".join(map(str, (


IanSudbery:
Hi, if you can add a very quick/simple test that would be great. Also, it seems that python 3.6 is no longer supported on the latest ubuntu through github actions. I've removed 3.6 from the testing on the master branch, but if you could do it on here (via .github/workflows/nosetests.yml) that would be great.

PR #576, opened 2023-02-24
check if gene_tag is empty

akmorrow13:
saving a flatfile will fail when `per_gene` is set and `gene_tag` is not set. This will occur when `per_gene` and `per_contig` are enabled: in this case, `gene_tag` cannot be set. In this case, `gene = read.get_tag(gene_tag)` will fail because `gene_tag is None`.


akmorrow13:
Can one of the admins verify this patch?

PR #586, opened 2023-03-03
Fix lack of help messages in 1.1.4

IanSudbery:
remove regression that caused umi_tools.py to quit before calling the command if --help was specified.

Added a test to test that there is help output from dedup. The reference will include the version number, which means this test will go out of date every time there is a version bump. On the other hand, should we test the help message from all tools?

### DIFF
--- a/tests/dedup_help.txt
+++ b/tests/dedup_help.txt
@@ -1,0 +1,135 @@
+UMI-Tools: Version 1.1.4
+
+dedup - Deduplicate reads using UMI and mapping coordinates
+
+Usage: umi_tools dedup [OPTIONS] [--stdin=IN_BAM] [--stdout=OUT_BAM]
+
+       note: If --stdout is ommited, standard out is output. To
+             generate a valid BAM file on standard out, please
+             redirect log with --log=LOGFILE or --log2stderr 
+
+For full UMI-tools documentation, see https://umi-tools.readthedocs.io/en/latest/
+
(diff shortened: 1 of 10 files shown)


IanSudbery:
@TomSmithCGAT This still needs a check if you can find time!

umi_tools/umi_tools.py
```diff
@@ -47,8 +47,6 @@ def main():
     elif len(argv) > 2 and  argv[2] in ["--help", "-h", "--help-extended"]:
         print("UMI-Tools: Version %s" % __version__)
```
TomSmithCGAT, line 48: Could add a (hidden) option to output help without the version number to resolve comment above?

tests/dedup_help.txt
```diff
@@ -0,0 +1,135 @@
+UMI-Tools: Version 1.1.4
```
TomSmithCGAT, line 1: If I was being really nitpicky, I'd say including the version number in the test file is going to be a pain as it means updating with every new version.

umi_tools/umi_tools.py
IanSudbery: or just only output the version number with `--version`

TomSmithCGAT: Yeah, I think that would be the best option

IanSudbery:
@TomSmithCGAT How about this?

PR #592, opened 2023-04-11
Changed yaml.load to yaml.safe_load

IanSudbery:
Recent debugging and testing revealed that changes to pyyaml meant changes were needed to how the testing sepcification was loaded.


TomSmithCGAT:
This is the same as https://github.com/CGATOxford/UMI-tools/commit/7d49256441abd06592c747f3da65de4d52554612, right?

Not even sure why the diff looks like it does above since the `safe_load` is already in the master branch 🤷‍♂️

TomSmithCGAT:
PR here: https://github.com/CGATOxford/UMI-tools/pull/542

Issue #555, opened 2022-08-18
bam dedup (umi_tools dedup)


zigenLi:
umi_tools dedup --extract-umi-method=tag --umi-tag=UB --cell-tag=CB \
	--method unique \
	-I input.bam -S output.bam \
	--log=LOGFILE

Hello, I used the above command to deduplicate the bam file, but there are still reads in the deduplicated bam file with the same coordinates and UMI

TomSmithCGAT:
Hi @zigenLi. Would you mind posting a few examples please. I suspect they may have different soft-clipping in the CIGAR string

zigenLi:
> 
example 1:
E100047948L1C007R01003932541	0	10	103122537	3	87M4D13M	*	0	0	GAATTTTGCGAAATGTTGTGACAGATATTTGTCAGAATTATTGGAAATATGTTATGGTGGATGGCATTTTTTCATATAAAATTATATTTCTGTTGAGACA	*	NH:i:2	HI:i:1	AS:i:86	nM:i:1	UR:Z:GTGCCCTCCA	CB:Z:045E30A27A	GN:Z:B2M	UB:Z:GTGCCCTCCA	RE:A:E	DB:Z:CELL1_N3
E100047948L1C010R02000347603	0	10	103122537	255	89M11S	*	0	0	GAATTTTGCAAAATGTTGTGACAGATATTTGTCAGAATTATTGGAAATATGTTATGGTGGATGGCATTTTTTCATATAAAATTATATTTCTGTTGTGTCA	*	NH:i:1	HI:i:1	AS:i:87	nM:i:0	UR:Z:GTGCCCTCCA	CB:Z:045E30A27A	GN:Z:B2M	UB:Z:GTGCCCTCCA	RE:A:E	DB:Z:CELL1_N3

example 2:
E100047948L1C042R02203017007	0	11	74998101	3	5S95M	*	0	0	CTATTTTTTTTTTTTTTTTAATGCTGAATATTGTGATTTCTTATTTGAGATTGAAAAAGACCCACTTGAAACTGATATATGTTAGAATATATTATGTTAA	*	NH:i:2	HI:i:1	AS:i:93	nM:i:0	UR:Z:TCGATGCAAG	CB:Z:7D43A87E2B	GN:Z:SF3B6	UB:Z:TCGATGCAAG	RE:A:E	DB:Z:CELL1_N3
[...]

PR #631, opened 2024-03-07
Add FAQ entry re identification of possible duplicates reads/pairs

TomSmithCGAT:
I've added two entries to the FAQ to explain how possible duplicate reads are identified and why the output of `dedup` can appear to contain reads with the same coordinates. I don't find this easy to explain clearly, so very happy to take suggested edits!

See #555 for an example motivation, though I feel like this has come up a few times in issues.


IanSudbery:
@TomSmithCGAT  Had a bit of a tweak (might have gotten carried away)

TomSmithCGAT:
Not got carried away at all - I think this is perfect! I had something very similar in mind, but was too lazy to do it!

I'm happy if we merge this now.

PR #632, opened 2024-03-07
Add issue templates

TomSmithCGAT:


TomSmithCGAT:
@IanSudbery - I think it would help to have issues templates. Happy to take feedback on suggested text, which is an edit of the default templates

TomSmithCGAT:
I'd really like to be able to include a 'banner' for the issues homepage, but that doesn't seem possible. 

I wonder if it might be helpful to include an issue template for 'advice' though, with a strong suggestion to put this sort of issue on a forum where help is likely to be more quickly obtained. I note that Mike Love has a never-closed issue on DESeq2 to that effect: https://github.com/thelovelab/DESeq2/issues/1

IanSudbery:
Happy to have these templates. 

You could have a "support request template" that just said "post to biostars", or is that a bit cheeky?

To be honest, the only person who ever answers umi_tools questions on biostars is me, so it doesn't make a massive amount of difference.

TomSmithCGAT:
Haha. No I think that would be fair enough, so long as it clarified when it's best to post to biostars. I'll add this.

Perhaps the more important point then is how easy it is for other users to find it. Closed issues aren't great for support/advice questions IMO.

TomSmithCGAT:
@IanSudbery - I've added a 'support request' template with links and suggestion to post to Biostars

TomSmithCGAT:
@IanSudbery  - You happy for me to merge this?

IanSudbery:
yes, go ahead

PR #640, opened 2024-04-23
change back to original matplotlib backend

sshen8:
When using the UMI-tools API only, I noticed it switches my Matplotlib backend to Agg, which messes with the plots I generate downstream. This PR changes the backend only when generating plots and then switches back to the user's original setting

### DIFF
--- a/umi_tools/whitelist_methods.py
+++ b/umi_tools/whitelist_methods.py
@@ -8,11 +8,11 @@ whitelist_methods.py - Methods for whitelisting cell barcodes
 import itertools
 import collections
 import matplotlib
+# change back to original user params after plotting
+_rcParams_orig = matplotlib.rcParams.copy()
 import copy
 import regex
 
-# require to run on systems with no X11
-matplotlib.use('Agg')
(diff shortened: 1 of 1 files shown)


TomSmithCGAT:
Thank you @sshen8. This looks like how I should have implemented it in the first place!

@IanSudbery - I'll merge this once tests pass.

umi_tools/whitelist_methods.py
```diff
@@ -100,6 +100,9 @@ def getKneeEstimateDensity(cell_barcode_counts,
 
     if plotfile_prefix:
 
+        # require to run on systems with no X11
```
IanSudbery, line 103: @sshen8 How would you feel about moving lines 11 and 12 to just here (line 102). This way, if the user changes any params after importing umi_tools, then those changes will get captured.

sshen8:
Yeah that's a good idea! I've just made the change, thanks for thinking of it

IanSudbery:
@TomSmithCGAT Happy once the tests complete.

PR #644, opened 2024-05-24
Update setup.py

IanSudbery:
Setup.py was still allowing install with python 2. This should force python to be at least 3 in order for setup.py to run.

I've also upped the required version of pysam. I don't know if the hack that was in place for pysam version still applies?


TomSmithCGAT:
All good to me. I'm also not sure if the hack is still required, but doesn't hurt to leave it in I guess 🤷‍♂️  Happy for you to merge

Issue #627, opened 2024-02-22
Compilation fails on python 3.12


s-andrews:
Using AlmaLinux 9.2 (RHEL9 derivative) with gcc 11.4.1 and python 3.12.2

Compilation using pip3 install umi-tools fails with:

```
      umi_tools/_dedup_umi.c: In function ‘__pyx_f_9umi_tools_10_dedup_umi_edit_distance’:
      umi_tools/_dedup_umi.c:1322:5: warning: ‘ma_version_tag’ is deprecated [-Wdeprecated-declarations]
       1322 |     __Pyx_GetModuleGlobalName(__pyx_t_4, __pyx_n_s_np); if (unlikely(!__pyx_t_4)) __PYX_ERR(0, 14, __pyx_L1_error)
            |     ^~~~~~~~~~~~~~~~~~~~~~~~~
      In file included from /bi/apps/python/3.12.2/include/python3.12/dictobject.h:90,
                       from /bi/apps/python/3.12.2/include/python3.12/Python.h:61,
                       from umi_tools/_dedup_umi.c:6:
      /bi/apps/python/3.12.2/include/python3.12/cpython/dictobject.h:22:34: note: declared here
         22 |     Py_DEPRECATED(3.12) uint64_t ma_version_tag;
            |                                  ^~~~~~~~~~~~~~
      umi_tools/_dedup_umi.c: In function ‘__Pyx_get_tp_dict_version’:
      umi_tools/_dedup_umi.c:1931:5: warning: ‘ma_version_tag’ is deprecated [-Wdeprecated-declarations]
       1931 |     return likely(dict) ? __PYX_GET_DICT_VERSION(dict) : 0;
            |     ^~~~~~
      In file included from /bi/apps/python/3.12.2/include/python3.12/dictobject.h:90,
                       from /bi/apps/python/3.12.2/include/python3.12/Python.h:61,
                       from umi_tools/_dedup_umi.c:6:
      /bi/apps/python/3.12.2/include/python3.12/cpython/dictobject.h:22:34: note: declared here
         22 |     Py_DEPRECATED(3.12) uint64_t ma_version_tag;
            |                                  ^~~~~~~~~~~~~~
      umi_tools/_dedup_umi.c: In function ‘__Pyx_get_object_dict_version’:
      umi_tools/_dedup_umi.c:1943:5: warning: ‘ma_version_tag’ is deprecated [-Wdeprecated-declarations]
[...]

Issue #362, opened 2019-08-14
dedup determinisitc - use a seed?


RichardCorbett:
Hi there,

When I run this command multiple times:
{code}
umi_tools dedup -I myBam.bam --extract-umi-method=tag --umi-tag=RX --read-length -S myBam.umi_tools.bam
{code}

I get the same number of reads returned each time, but the exact reads returned each time are different.  Is there a way to get the same reads with each run?

thanks
RIchard

IanSudbery:
To use a seed with `umi-tools`, use the `--random-seed=` option. 

You might also find you need to set the `PYTHONHASHSEED=` environment variable as well to make sure hashes are returned in the same order, although I think the sorts should take care of that these days....

This is how we run our tests:

```
#!/use/bin/env bash
export PYTHONHASHSEED=0
umi_tools dedup --random-seed=123456789 .......
```

RichardCorbett:
Perfect. Thanks.  Sorry I somehow missed this myself.

IanSudbery:
I'm closing this. Please reopen if you need more advice.

PR #655, opened 2024-07-31
Update testing suite to pytest

eachanjohnson:
In running tests for PR #654, I ran into an issue with testing on my python3.11 platform. The nose package itself was running into import errors, and on further digging does not seem to be maintained any longer.

So I've lightly rewritten tests/test_style.py and tests/test_umi_tools.py to use pytest. 

The tests now run for python3.11, so I expanded the range of Python versions in GitHub actions. Unfortunately, `umi_tools` doesn't compile under python3.12, which I think is a known issue #627.

Also, not all the tests run successfully, which I guess is related to the non-deterministic issue #362 and PRs #550, #98.

This may not be the kind of contribution you're looking for, but I ended up doing the work so I thought I'd hand it over 😄


TomSmithCGAT:
Thank you for this @eachanjohnson! You're right that `nose` is no longer maintained. Moving into a different testing framework has been on the to-do list for a while now. I really appreciate you taking the time to do this for us 🙏 

I've confimed the 18 failed tests with python 3.11 outside of the github actions too. Given past experiences and looking into a few of the failures, I'm assuming this is due to changes in how dictionaries work between py3.10-3.11, but I can't see anything in the release notes relating to this. @IanSudbery - Do you have any idea what's going on here?

IanSudbery:
So I had a quick look at one test. There seems to be a couple of things going on here. One is the sort order has gotten out of whack -

Taking the first failed test:
```
umi_tools dedup --paired --log=test.log --filter-umi --umi-whitelist=tests/umi_whitelist.tsv --umi-whitelist-paired=tests/umi_whitelist.tsv --out-sam
--random-seed=123456789 -I tests/whitelist_umi_input.bam
```

There are 2866 differences:

```
$ diff -y --suppress-common-lines test.out tests/whitelist_umi_output.sam | wc -l
2232
```

However, if we sort there are only 3:

```
$ diff -y --suppress-common-lines <(sort test.out) <(sort tests/whitelist_umi_output.sam) | wc -l
3
```

My guess is that this must be a change in how pysam (and possibly htslib/samtools) sorts bam files. 

There are still three lines different though:

```
$ diff -y --suppress-common-lines <(sort test.out) <(sort tests/whitelist_umi_output.sam)
NS500105:308:HCGGHAFXY:1:11203:7544:2472_TACGAACCGTGTACTG     <
NS500105:308:HCGGHAFXY:1:21106:2617:15679_GTTCACCTAGCTCTAG    <
NS500105:308:HCGGHAFXY:1:21106:2617:15679_GTTCACCTAGCTCTAG    <
```
Three extra reads that arn't in the reference.

IanSudbery:
The sort order issue can probably be sorted by add more sort directives in the tests. I believe the remaining issues are caused by changes in which way ties are being broken when a pseudo-random choice between two equivalent UMIs is being made. At the moment, I think we just choose whichever comes out top from `sorted`, which in turn will depend on what order they came out of the original dictionary.  

One way in which this might have changed is that the default hashing algorithm for `str` changed between 3.10 and 3.11. 

@TomSmithCGAT  This leaves us in the somewhat difficult position of trying to decide whether to implement different tests for different versions, or finally chase down dealing with getting something predictable out of extracting things from a dict. 

One alternative would be to explicitly check whether the top choice is a tie, and use an explict random choice to pick. 

I don't know how much slow down this would give.

TomSmithCGAT:
I think we can just merge #550 and be done with the non-deterministic issues, right? 

For some reason, I thought that was waiting on you running some benchmarks, but I see that was done a full year ago 😱. If you're still happy with the changes in the code in #550, I'll resolve the conflicts and update the test files so we can merge that PR

IanSudbery:
@TomSmithCGAT If you had time to do that, that would be great.

TomSmithCGAT:
Hi @eachanjohnson. I finally got around to merging #550, which should resolve the non-deterministic issue causing issues here. I tried to make push commits to this branch so it will pass the tests but I got a permission denied error. It's just a few steps:

1. Pull master branch in your fork
2. Merge master into switch-to-pytest branch
3. That'll give a conflict error for `.github/workflows/tests.yml`. You can resolve that in favour of your branch and remove the line with `export PYTHONHASHSEED=0` under the run directive, since that's no longer needed. 

After that, you should see that the tests pass regardless of whether `PYTHONHASHSEED` is set. Hopefully that'll resolve the issues with python 3.10/3.11 too. Let's see..

eachanjohnson:
Thanks, @TomSmithCGAT! Feeling ambitious, I added python3.12 to the tests, which also passed 🚀

IanSudbery:
@eachanjohnson Thats for doing that. @TomSmithCGAT Are we good to merge?

TomSmithCGAT:
Yes. All good. I'll do that now.

Thank you again @eachanjohnson 🙏

PR #657, opened 2024-08-07
Rebuilt _dedup_umi.c using cython 3.0.11

IanSudbery:
Compilation was failing on python 3.12. Rebuilt _dedup_umi.c from _dedup_umi.pyx using newer version of cython.


TomSmithCGAT:
I'm going to have to take it on trust that the new `_dedup_umi.c` is correct 😉 . Have you checked this enables compiling with python 3.12? If so, merge away from my perspective!

TomSmithCGAT (approved):

PR #658, opened 2024-08-09
Update Python versions for tests

sebastian-luna-valero:
See: https://devguide.python.org/versions/

xrefs:
* https://github.com/CGATOxford/UMI-tools/pull/657
* https://github.com/CGATOxford/UMI-tools/issues/627

Once these tests work, we could re-enable python 3.12 in the [bioconda recipe](https://github.com/bioconda/bioconda-recipes/blob/e5e9d1964cc129c691e5ea510f6d16cf30b394eb/recipes/umi_tools/meta.yaml#L16-L17).


sebastian-luna-valero:
ok, so I am in a rush and don't have time to fix these errors.

we should replace `nose` with `nose2` or `pytest`. However, that means refactoring `tests/test_umi_tools.py`.

could you please help here?

IanSudbery:
Hi @sebastian-luna-valero. There is already work on this at #655. Different python versions result in different results from the tests, so we are waiting on #550 which should make the results more deterministic as long as a random seed is provided.

TomSmithCGAT:
Now resolved by #655. Thanks @sebastian-luna-valero

PR #672, opened 2024-12-20
Update QUICK_START.md

user-tq:
correct means of ‘N‘ and ‘X‘


IanSudbery:
The original is correct, and the suggestion here is incorrect. Please see documentation at https://umi-tools.readthedocs.io/en/latest/regex.html. This confusion is behind the difficulty you are having with #673 .

PR #677, opened 2025-02-15
Update _dedup_umi.pyx

6x7p3:
UMI_tools doesn't run on my linux with the following error:
AttributeError: `np.Inf` was removed in the NumPy 2.0 release. Use `np.inf` instead.

replacing np.Inf fixed the problem

### DIFF
--- a/umi_tools/_dedup_umi.pyx
+++ b/umi_tools/_dedup_umi.pyx
@@ -11,7 +11,7 @@ cpdef int edit_distance(a, b):
     # we only want to define hamming distances between barcodes with the same length 
     lb = len(b)
     if la != lb:
-        return np.Inf
+        return np.inf
 	    
     for k from 0 <= k < la:
         if aa[k] != bb[k]:


TomSmithCGAT:
Thank you @6x7p3 🙏

IanSudbery:
Do we know when `np.inf` was introduced? Do we need to change the numpy pin?

TomSmithCGAT:
Not exactly sure. I thought it was long ago not too worry about, but I hadn't appreciated our numpy requirement is only >1.7 (2013!). In practice, nobody will end up using a version that old due to other dependencies obviously. We're also We're already using `np.inf` here: https://github.com/CGATOxford/UMI-tools/blob/ee39b777ee69505c519bc6aa0d988a436b28616a/umi_tools/whitelist_methods.py#L429

The first instance I can find of 'np.inf' in release notes is 1.16.4: https://numpy.org/doc/2.0/release/1.16.4-notes.html

The first time it appears in the docs seems to be 1.15: https://numpy.org/doc/1.15/reference/constants.html?highlight=inf#numpy.inf

So I guess we could update to `numpy>1.15`

IanSudbery:
Seems reasonable. I tried installing 1.7 to test, and the pacakge seems to be broken anyway.

Issue #688, opened 2025-07-09
umi_tools extract inserts DOS line endings into output


j-miko:
**Describe the bug**
umi_tools extract output has DOS style (CR LF) line endings, they seem mixed with Unix style LF and seem to appear mostly at the end of each FASTQ entry. This causes downstream steps to throw unexpected end of file errors.

**To Reproduce**
umi_tools extract; same thing happens when reading reads from file or piping them through stdin; whether input/output is compressed doesn't seem to make a difference as well.

**Expected behavior**
Enforce Unix-style LF

**Environment**
 - OS: Linux 
 - UMI-tools version: 1.1.6
 - How was UMI-tools installed: conda

**Additional context**
umi_tools extract run as part of Snakemake pipeline with default options

j-miko:
So in the end it all seems to be a bit more convoluted than I initially thought:

- some raw FASTQs do indeed have DOS endings - a nice lesson for me to always check raw files from collaborators...
- the commit above sanitizes them nicely and output file have Unix endings only
- the downstream issue does not seem to be dependent on this at all: we're using flexbar in this specific pipeline and it keeps throwing an 'unexpected end of input' error whether the input files are coming from the patched version of umi_tools, the old one, sanitized manually using `tr -d '\r'` or input raw without any modifications. I should probably move this to flexbar support though.

IanSudbery:
Closing this issue as its been merged for nearly a year with no activity.

PR #687, opened 2025-06-13
RTD fixes ahead of new documentation

IanSudbery:
RTD now requires a .readthedocs.yml (seems it has for several years). Added this file, and also fixed the docs/requirements.txt and conf.py so that imports are mocked rather than having to install dependencies to build the documents.

PR #689, opened 2025-07-10
Strip fastq lines, rather than explicitly removing last char. 

IanSudbery:
This should help if input file has DOS-like line endings, replacing them with UNIX ones.


IanSudbery:
@TomSmithCGAT Given what you said on #687 and given you've chipped in on #688 I'm going to just go ahead and merge this.

PR #696, opened 2026-01-24
{is} cram support

IanSudbery:
Support CRAM files as input and output for `dedup`, `group`, `count` and `prepare-for-rsem`

Foramts should be auto detected from filenames, fall back on `--in-format`/`--out-format`.

Reference location specified by `--reference-filename`. Can sometimes be guessed for inputs (from the file header, or the reference cache), but is neccessary for writing. If missed out when writing, the reference will be embbedded in the file and a warning raised. 

Currently no tests. 

I have fixed the testing framework though, so as to deal with #695.


IanSudbery:
Tests now added.

PR #700, opened 2026-04-10
fix: rename build backend module to use underscore instead of hyphen

HSiga:
## Problem

The hyphen in `setuptools-cython` is an invalid Python identifier.
Tools like `uv` and `pixi` generate an import statement at build time that causes a hard SyntaxError:

```
      [stderr]
        File "<string>", line 8
          import setuptools-cython as backend
                           ^
      SyntaxError: invalid syntax
```

## Fix

- Renamed `setuptools-cython/setuptools-cython.py` to `setuptools_cython.py`
- Updated `build-backend` in `pyproject.toml` from `"setuptools-cython"` to `"setuptools_cython"`

The `backend-path` pointing to the `setuptools-cython/` directory is unchanged.

## Tested

Confirmed the SyntaxError reproduces in a clean venv before the fix, and
that `uv pip install .` succeeds after.


IanSudbery (approved):
approve subject to tests.

IanSudbery:
@HSiga  Thanks for this. Sorry for the delay merging.

PR #712, opened 2026-06-16
Fix/output stats performance

TomSmithCGAT:
`--output-stats` can be very slow on large BAMs due to two compounding costs:

1. **An extra full BAM scan at startup.** `random_read_generator` parsed the entire BAM
   before deduplication began, solely to build the UMI frequency distribution needed for
   null sampling.

2. **No way to trade statistical resolution for speed.** Edit distance statistics were always
   computed for every position in the BAM, with no option to subsample. The only option was to run separately with `--chrom` as a crude subsample.

### Changes

**Removed extra BAM scan**

The UMI frequency distribution for null sampling is now built from data already collected
during the main deduplication loop, rather than a separate upfront BAM scan. The distribution
covers all positions where at least one read survives deduplication — equivalent to the old
approach in normal usage. Single-UMI positions are also short-circuited directly to `-1`
without a null draw, since the result is known by convention. `random_read_generator` and its
unused `pysam` import have been removed from `umi_methods.py`. Statistical equivalence of the
old and new null distributions was confirmed across 100 random seeds (Mann-Whitney U, all
p > 0.05).

**New `--stats-sample-fraction` option**

Controls the fraction of positions used for edit distance computation (default `1.0`). At
`0.1`, only 10% of positions contribute to the edit distance histogram, reducing that cost by
~90%. The `_per_umi.tsv` and `_per_umi_per_position.tsv` outputs are unaffected — they are
always computed from all positions.

### Docs

The FAQ and quick-start guide are updated to mention `--stats-sample-fraction` and `--chrom` as options for reducing run time when using `--output-stats`.

### Testing

**Why test reference files were updated**

The null edit-distance values in `tests/single_stats_py3_edit_distance.tsv` changed as a
result of this PR, despite the underlying UMI frequency distribution used for null
sampling is equivalent. In the original code, `random_read_generator` pre-scanned the BAM
immediately after `np.random.seed()` was called, so the first random draws from that seed
went into building the null sampling buffer. In the new code the seed is consumed differently:
the main loop runs first (using `np.random.random()` for sampling-fraction checks at each
position), and null draws happen in a post-loop batch starting from a fresh re-seed. Because
the sequence of random numbers drawn before the null values are sampled differs, the same seed
produces different null values. The observed (non-null) columns are identical between old and
new code.

**Statistical equivalence of old and new null distributions**

To confirm that the change in draw order does not alter the statistical properties of the null
distribution, both old and new code were run across 100 random seeds on the test BAM
(`chr19.bam`, `--method=cluster`). The `unique_null` counts at each edit distance bin were
compared using the Mann-Whitney U test. All p-values were well above 0.05 (range 0.05–1.00),
confirming no statistically significant difference between the old and new null distributions.

umi_tools_null_old_vs_new


TomSmithCGAT:
@IanSudbery - I don't have ready access to a more complex BAM file. Would you have something you could share with me?

TomSmithCGAT:
I've run some benchmarks on a larger file, albeit not a complex one:

  - 4,342,982 input read pairs
  - 4,152,669 reads out (deduplicated)
  - 4,142,656 positions deduplicated
  - Mean unique UMIs/position: 1.00 — the vast majority of positions have just one UMI
  - Max unique UMIs/position: 128

### 1. master vs fix — `--output-stats`, fraction=1.0

  | Branch | N | Time, mean ± SD (s) | RSS, mean ± SD (MB) |
  | --- | --- | --- | --- |
  | master | 5 | 180 ± 3 | 2053 ± 70 |
  | fix | 5 | 154 ± 2 | 1700 ± 89 |

  Fix branch is 26s faster (-14%) and uses 353 MB less memory (-17%).
  
  ### 2. No-stats baseline (1 run each)

  | Branch | Time (s) | RSS (MB) |
  | --- | --- | --- |
  | master | 116 | 1299 |
  | fix | 117 | 1336 |

  Both branches are identical without stats — all gains are in the stats code path.

  ### 3. fix branch — `--stats-sample-fraction` (n=3 each)
  
  | Fraction | Time, mean ± SD (s) | RSS, mean ± SD (MB) |
  | --- | --- | --- |
  | 0.1 | 151 ± 3 | 1524 ± 30 |
  | 0.25 | 153 ± 3 | 1666 ± 71 |
  | 0.5 | 152 ± 3 | 1718 ± 111 |
  | 1.0 | 156 ± 2 | 1687 ± 13 |
  | no-stats | 117 | 1336 |

  Subsampling fraction has minimal wall-time impact — BAM I/O dominates and the edit distance computation is no longer the bottleneck after the buffered null-draw fix.

  ### 4. fix branch — `--per-umi-stats` flag (n=3 each)

  | Condition | Time, mean ± SD (s) | RSS, mean ± SD (MB) |
  | --- | --- | --- |
  | with --per-umi-stats | 152 ± 1 | 1679 ± 37 |
  | without --per-umi-stats (default) | 126 ± 1 | 1789 ± 73 |
  | no-stats baseline | 117 | 1336 |

  Making `_per_umi.tsv` opt-in saves ~26s and brings the default stats run to within 9s of the no-stats baseline. The memory difference is noise — peak RSS is dominated by the read buffer, not the stats structures.

TomSmithCGAT:
@IanSudbery - As I understand it, you want to make bugfix release (v1.1.7) before a feature release (v1.2). I'll leave this here until you're ready to merge it in.

TomSmithCGAT:
Testing on a different BAM file which is larger and slightly more complex is illuminating:

Reads: Input Reads: 39829499
Number of reads out: 5505460
Total number of positions deduplicated: 3325859
Mean number of unique UMIs per position: 2.03
Max. number of unique UMIs per position: 421

  | Condition | Mean (s) | SD (s) | vs master |
  |---|---|---|---|
  | new (no stats) | 167 | 2.1 | −29% |
  | new --output-stats f=0.1 | 178 | 2.2 | −24% |
  | new --output-stats f=0.5 | 184 | 1.6 | −22% |
  | new --output-stats | 192 | 0.8 | −18% |
  | new --output-stats --per-umi-stats | 193 | 2.3 | −18% |
  | master --output-stats | 235 | 3.7 | — |


Figure_1


In this case, dropping the `--per-umi-stats` gives almost no benefit but `--stats-sample-fraction=0.1` gets pretty close to not running stats at all. 

On reflection, I think perhaps the best option is to output per-umi stats by default and include a flag to drop it, with explanation in the docs to explain the options to deal with long stats run time. I'll do that shortly

TomSmithCGAT:
Done. Ready for merge whenever's best. Suggested release notes for these changes:

### dedup --output-stats performance improvements
  
  The time taken to run umi_tools dedup --output-stats has been significantly reduced (~18% faster on a 44.9M read BAM):

  - Single-UMI positions are no longer subjected to an edit distance computation, as their average distance is trivially known
  - The UMI frequency distribution used for null model sampling is now built from data collected during the main dedup loop, removing the extra BAM scan that previously occurred at startup
  - Null UMI draws are now buffered (100k UMIs drawn per call and sliced per position), reducing the overhead of repeated numpy calls

### New options for dedup --output-stats

  - --stats-sample-fraction — compute edit distance statistics on a random subset of positions (e.g. --stats-sample-fraction=0.1 for 10%). The _per_umi_per_position.tsv output is unaffected and always uses all
  positions. Useful for very large datasets where the edit distance null distribution is the bottleneck.
  - --no-per-umi-stats — suppress the per-UMI summary table (_per_umi.tsv). This table is now written by default when --output-stats is used. Pass --no-per-umi-stats if it is not needed and the dataset has many
  distinct UMI sequences (e.g. long UMIs or high sequencing depth), as the aggregation step can be slow in these cases.

